PR #1, opened 2018-05-15
Read and display countdown signs from an external URL

kohlerjp:
Bare-bones initial implementation.

Reads a JSON list of signs from an external URL and displays them on an index page.
On Startup, the Signs State reads in the json file and stores it as it's internal state. Loading the index page will fetch and display the list of signs from the GenServer

Example: 
_signs_list


lib/signs_ui/signs/sign.ex
```diff
@@ -0,0 +1,13 @@
+defmodule SignsUI.Signs.Sign do
+  defstruct [
+    id: "",
+    disabled?: false
```
sghiaa: i thought we were going with enabled?

PR #2, opened 2018-05-15
Keep internal state on updates

kohlerjp:
Another piece of [L: PIO can edit the contents of the S3 on/off file using a UI](https://app.asana.com/0/584764604969369/629803768071611)

This turns the list of signs into a form. Posting the form will update the internal state of the GenServer, so subsequent page loads will be up to date.

### DIFF
--- a/lib/signs_ui/signs/sign.ex
+++ b/lib/signs_ui/signs/sign.ex
@@ -10,4 +10,9 @@ defmodule SignsUI.Signs.Sign do
       enabled?: Map.get(values, "enabled")
     }
   end
+
+  def update_enabled(sign, enabled_string) do
+    enabled? = enabled_string == "true"
+    %{sign | enabled?: enabled?}
+  end
 end
--- a/lib/signs_ui/signs/sign.ex
+++ b/lib/signs_ui/signs/sign.ex
@@ -10,9 +10,4 @@ defmodule SignsUI.Signs.Sign do
       enabled?: Map.get(values, "enabled")
     }
   end
-
-  def update_enabled(sign, enabled_string) do
-    enabled? = enabled_string == "true"
-    %{sign | enabled?: enabled?}
-  end
 end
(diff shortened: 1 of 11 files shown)


lib/signs_ui_web/templates/signs/index.html.eex
```diff
@@ -1,8 +1,12 @@
 <h3>Countdown Signs</h3>
 <div>
-  <ul>
+  <%= form_for @conn, signs_path(@conn, :update), [as: :signs], fn f -> %>
     <%= for {sign_id, sign} <- @signs do %>
-    <li><%= sign_id%> : <%= if sign.enabled?, do: "Enabled", else: "Disabled" %></li>
+      <div>
+        <label><%= sign_id %></label>
```
losvedir: You should make it so that clicking on the label will select / deselect the checkbox. I usually do this by just putting the checkbox _within_ the label tag, but you can also go the `for=` route.

lib/signs_ui/signs/sign.ex
```diff
@@ -10,4 +10,9 @@ defmodule SignsUI.Signs.Sign do
       enabled?: Map.get(values, "enabled")
     }
   end
+
+  def update_enabled(sign, enabled_string) do
+    enabled? = enabled_string == "true"
```
losvedir: I generally think of this as the responsibility of the controller. That's the boundary where external input is validated and converted into the stuff that the app uses internally. Passing around strings like `"true"` is kind of a code smell, to me.

lib/signs_ui_web/templates/signs/index.html.eex
```diff
@@ -1,8 +1,12 @@
 <h3>Countdown Signs</h3>
 <div>
-  <ul>
+  <%= form_for @conn, signs_path(@conn, :update), [as: :signs], fn f -> %>
     <%= for {sign_id, sign} <- @signs do %>
-    <li><%= sign_id%> : <%= if sign.enabled?, do: "Enabled", else: "Disabled" %></li>
+      <div>
+        <label for=<%= sign_id %>><%= sign_id %></label>
```
losvedir: Little thing, but can you actually switch the order here, so all the checkboxes line up?

kohlerjp: I was skeptical but that looks much better

PR #3, opened 2018-05-16
Post to AWS

kohlerjp:
Final piece of [L: PIO can edit the contents of the S3 on/off file using a UI](https://app.asana.com/0/584764604969369/629803768071611)

Mostly adds in the `ex_aws` dependency and uses that to get the current config file from AWS, and also to post the updated JSON after the Submit button is clicked


lib/signs_ui/signs/request.ex
```diff
@@ -1,28 +1,21 @@
...
-        Logger.warn("Could not load signs. Response returned with status code #{inspect status}")
-        %{}
-      {:error, %HTTPoison.Error{reason: reason}} ->
-        Logger.warn("Could not load signs: #{inspect reason}")
+  def get_signs() do
+    case S3.get_object() do
+      {:ok, %{body: json}} ->
+        parse(json)
+      {:error, reason} ->
+        Logger.warn("Could not load Signs config from AWS: #{inspect reason}")
```
losvedir: The app is kind of useless if this fails. (Just ran into this deploying this branch to AWS - had permissions wrong and was consistently shown a blank page with no signs).

I think it's better to log an error and then return some sort of error tuple here, so that the `GenServer.init` that calls this can return `{:stop, :failed_to_load_config}`.

That will in turn cause the app to fail to start, and work out a little better in the AWS environment. For instance, ElasticBeanstalk, when you deploy a new version, will boot up the new one in the background, and if all is well, switch to it. By making it fail to start up here, it will go on using the previous version (which presumably still works).

losvedir: Same in the other branch if it loads but fails to parse, I guess.

lib/signs_ui/signs/state.ex
```diff
@@ -20,16 +20,19 @@ defmodule SignsUI.Signs.State do
...
   def handle_call(:get_all, _from, signs) do
     {:reply, signs, signs}
   end
   def handle_call({:update, enabled_values}, _from, signs) do
+    external_post_mod = Application.get_env(:signs_ui, :signs_external_post_mod)
     updated_signs = Signs.Signs.update_enabled_flags(enabled_values, signs)
-    {:reply, updated_signs, updated_signs}
+    {status, new_sign_state} =  updated_signs
+                                |> external_post_mod.update()
+                                |> external_post_mod.handle_response(signs, updated_signs)
```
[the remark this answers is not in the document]
kohlerjp: Yeah that makes sense

lib/signs_ui_web/controllers/signs_controller.ex
```diff
@@ -7,17 +7,28 @@ defmodule SignsUiWeb.SignsController do
...
 
   def update(conn, params) do
-    updated_signs = params
-                    |> get_enabled_map()
-                    |> SignsUI.Signs.State.update()
-    conn
-    |> put_flash(:success, "Signs updated successfully")
-    |> render("index.html", signs: updated_signs)
+    conn = perform_update(conn, params)
+    redirect(conn, to: "/")
```
losvedir: The usual flow is redirect on success, render on failure. The advantage of that approach is that if it fails for some reason (maybe a blip in S3), the app state won't change, but the _form_ at least will keep the changes they made. If the error is something like "please try again", then all they need to do is click submit again.

kohlerjp: Yeah, I wasn't sure if it'd be confusing if it looked the changes went through. But it would be annoying to reset all the checkboxes. Hopefully the message should make it clear

lib/signs_ui/signs/state.ex
```diff
@@ -20,16 +20,25 @@ defmodule SignsUI.Signs.State do
   end
 
   def init(_) do
-    signs_url = Application.get_env(:signs_ui, :signs_url)
-    signs = Signs.Request.get_signs(signs_url)
-    {:ok, signs}
+    Signs.Request.get_signs()
```
losvedir: `get_signs/0` here returns `{:ok, any()} | {:error, any()}` , which is right since it can either succeed or fail.

However, the GenServer init callback has a [different spec](https://hexdocs.pm/elixir/GenServer.html#c:init/1). If it returns `{:ok, state}` then the GenServer starts up with that initial state. But if you want to signal that the GenServer failed to start, you need to return `{:stop, reason}`.

I think it's correct for `get_signs` to return `{:error, ...}`, you should just handle it here and return either `{:ok, ...}` or `{:stop, ...}`.

PR #9, opened 2018-05-23
Use bootstrap grid with column headers

kohlerjp:
Asana Ticket: [Add columns for "Status" and "Station"](https://app.asana.com/0/584764604969369/683023916934070)

I think `Sign` might be a better header than `Station`, what do y'all think?

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -1,1 +1,2 @@
 /* This file is for your main application css. */
+@import  'signs.css'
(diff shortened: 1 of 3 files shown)


sghiaa:
screen shot 2018-05-23 at 2 39 04 pm
looks a little weird at md/sm

sghiaa:
also i agree about sign vs station

PR #10, opened 2018-05-23
add .tool-versions

dmaltzan:


.tool-versions
```diff
@@ -0,0 +1,2 @@
+erlang 20.1
+elixir 1.5.1
```
kohlerjp: I've been running this with elixir 1.6.4. I don't use `.tool-versions` when I run locally, so I'm not sure how big of a deal that is.

dmaltzan: updated

PR #11, opened 2018-05-23
add username/pw

dmaltzan:
is this a good idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ To start your Phoenix server:
 
   * Install dependencies with `mix deps.get`
   * Install Node.js dependencies with `cd assets && npm install`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ To start your Phoenix server:
 
   * Install dependencies with `mix deps.get`
   * Install Node.js dependencies with `cd assets && npm install`
-  * Start Phoenix endpoint with `mix phx.server`
+  * Start Phoenix endpoint with `USERNAME=username PASSWORD=password mix phx.server` (substituting whatever username/password you want)
```
kohlerjp: Yeah, I don't really see the harm in this. Alternatively, we could just mention that `USERNAME` and `PASSWORD` are environment variables. Also maybe mention that it's run via `mix run --no-halt`

PR #14, opened 2018-05-29
 User friendly names for sign zones, and ordering by system layout

sghiaa:
[[S] User friendly names for sign zones, and ordering by system layout](https://app.asana.com/0/584764604969369/683023916934071)

this is mattapan and chelsea - we can add more as we add them i suppose

### DIFF
--- a/lib/signs_ui_web/templates/signs/index.html.eex
+++ b/lib/signs_ui_web/templates/signs/index.html.eex
@@ -9,11 +9,109 @@
         <b>Sign</b>
       </div>
     </div>
-    <%= for {sign_id, sign} <- @signs do %>
-      <%= checkbox(f, sign_id, value: sign.enabled?, id: sign_id, class: "col-md-1 col-xs-2 pull-right ios8-switch") %>
-      <%= content_tag :label, sign_id,
-        for: sign_id,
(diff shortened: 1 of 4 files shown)


lib/signs_ui_web/templates/signs/index.html.eex
```diff
@@ -9,11 +9,109 @@
         <b>Sign</b>
       </div>
     </div>
-    <%= for {sign_id, sign} <- @signs do %>
-      <%= checkbox(f, sign_id, value: sign.enabled?, id: sign_id, class: "col-md-1 col-xs-2 pull-right ios8-switch") %>
-      <%= content_tag :label, sign_id,
-        for: sign_id,
-        class: if !sign.enabled?, do: "text-muted col-md-11 col-xs-10", else: "col-md-11 col-xs-10" %>
+    <%= if @signs["ashmont_mezzanine"] do %>
```
kohlerjp: This whole thing should probably be a view function. There's really no need to duplicate it for every sign. That way we can end up doing something like:

```
<%= for {id, label} in [{"ashmont_mezz", "Ashmont Mezzanine"}, {"cedar_grove_outbound", "Cedar Grove Outbound"}, ...] %> do
  <%= switch_for_sign(id, label) %>
<% end %>
```

The list of signs / labels could even be created at the controller level.

PR #18, opened 2018-06-08
Route Navigation Page

kohlerjp:
Asana Ticket: [Create parent navigational page for signs control app](https://app.asana.com/0/584764604969369/684786859866691)

Separates signs by route. Gives each route it's own page.

### DIFF
--- a/lib/signs_ui/mock/aws_request.ex
+++ b/lib/signs_ui/mock/aws_request.ex
@@ -8,7 +8,11 @@ defmodule SignsUI.Mock.AwsRequest do
       "cedar_grove_outbound" => %{"enabled" => "true"},
       "cedar_grove_inbound" => %{"enabled" => "true"},
       "butler_outbound" => %{"enabled" => "true"},
-      "butler_inbound" => %{"enabled" => "true"}
+      "butler_inbound" => %{"enabled" => "true"},
+      "maverick_eastbound" => %{"enabled" => "true"},
+      "maverick_westbound" => %{"enabled" => "true"},
+      "eastern_ave_inbound" => %{"enabled" => "true"},
+      "eastern_ave_outbound" => %{"enabled" => "true"}
     }
     {:ok, %{body: Poison.encode!(signs)}}
   end
--- a/lib/signs_ui/signs/names.ex
+++ b/lib/signs_ui/signs/names.ex
@@ -1,0 +1,87 @@
+defmodule SignsUI.Signs.Names do
+
+  def sign_routes() do
+    %{
+      mattapan: "Mattapan",
+      silver_line: "Silver Line",
+      blue: "Blue Line"
+    }
+  end
+
+  def mattapan() do
+    [
+      {"ashmont_mezzanine", "Ashmont Mezzanine"},
+      {"cedar_grove_outbound", "Cedar Grove Outbound"},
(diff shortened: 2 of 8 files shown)


lib/signs_ui/mock/aws_request.ex
```diff
@@ -8,7 +8,11 @@ defmodule SignsUI.Mock.AwsRequest do
       "cedar_grove_outbound" => %{"enabled" => "true"},
       "cedar_grove_inbound" => %{"enabled" => "true"},
       "butler_outbound" => %{"enabled" => "true"},
-      "butler_inbound" => %{"enabled" => "true"}
+      "butler_inbound" => %{"enabled" => "true"},
+      "maverick_eastbound" => %{"enabled" => "true"},
+      "maverick_westbound" => %{"enabled" => "true"},
+      "eastern_ave_inbound" => %{"enabled" => "true"},
+      "eastern_ave_outbound" => %{"enabled" => "true"}
```
losvedir: Trailing comma, please! If we had one before we wouldn't have this unnecessary red line above in the diff here!

lib/signs_ui/signs/names.ex
```diff
@@ -0,0 +1,87 @@
+defmodule SignsUI.Signs.Names do
+
+  def sign_routes() do
+    %{
+      mattapan: "Mattapan",
+      silver_line: "Silver Line",
```
losvedir: For consistency with the other two can this just be `silver`?

lib/signs_ui_web/controllers/signs_controller.ex
```diff
@@ -15,71 +21,14 @@ defmodule SignsUiWeb.SignsController do
...
-      {"government_center_mezzanine_eastbound", "Government Center Mezzanine Eastbound"},
-      {"government_center_westbound", "Government Center Westbound"},
-      {"bowdoin_eastbound", "Bowdoin Eastbound"},
-      {"bowdoin_mezzanine", "Bowdoin Mezzanine"}
-    ]
+  def sign_names(%{"route" => "mattapan"}), do: Signs.Names.mattapan()
+  def sign_names(%{"route" => "blue"}), do: Signs.Names.blue_line()
+  def sign_names(%{"route" => "silver_line"}), do: Signs.Names.silver_line()
+  def sign_names(%{"route" => "all"}) do
+    Signs.Names.mattapan() ++ Signs.Names.silver_line ++ Signs.Names.blue_line()
```
losvedir: For ordering, how about subway before bus, and then subway ordered the way it is on mbta.com? I believe the proper order is specified via the API, but I don't think we need to reach out for that here... it's unlikely to change at this point.

lib/signs_ui_web/controllers/signs_controller.ex
```diff
@@ -108,6 +57,10 @@ defmodule SignsUiWeb.SignsController do
   end
 
   defp do_params_to_signs({sign_id, sign}, enabled_map) do
-    {sign_id, %{sign | enabled?: Map.get(enabled_map, sign_id)}}
```
losvedir: What was this change for?

kohlerjp: All the signs that weren't on the individual route page were getting `nil` set for this value, since we're still passing all the signs to the controller.

PR #19, opened 2018-06-11
S: Add a line-specific title to the child page (e.g. "Mattapan Realtime Signs")

sghiaa:
[S: Add a line-specific title to the child page (e.g. "Mattapan Realtime Signs")](https://app.asana.com/0/584764604969369/705930618035017)

### DIFF
--- a/lib/signs_ui_web/templates/signs/index.html.eex
+++ b/lib/signs_ui_web/templates/signs/index.html.eex
@@ -1,4 +1,4 @@
-<h3>Countdown Signs</h3>
+<h3><%= @route_name %> Signs</h3>
 <div class="container signs-container">
   <%= form_for @conn, signs_path(@conn, :update), [as: :signs, class: "signs-form"], fn f -> %>
     <div class="row signs-headers-row">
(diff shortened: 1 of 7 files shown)


lib/signs_ui_web/templates/signs/index.html.eex
```diff
@@ -1,4 +1,3 @@
-<h3>Countdown Signs</h3>
```
losvedir: I think we should *keep* this one, and get rid of the "MBTA Realtime Signs" header above, rather than vice versa. Since the param is a property of _this_ controller, it seems like the route should be a property of this template. Presumably, if the app grew to have other controllers or concerns, we wouldn't want the route header up in the top level layout template.

sghiaa: I kind of agree and thats how i had it originally, but the mockup has the main header being the one thats replaced so i tried to match that.

sghiaa: if you look at the previous commit, that is how it was - would you rather we went with that approach?

losvedir: I see... Well, if we get rid of the other one entirely, can we make this H3 and H1 and therefore be true to the mockup? I don't think there's usually headers on layout templates, so it'd be fine to put the H1 in the view template.

sghiaa: ok, thats fine with me

PR #24, opened 2018-06-15
toggle all buttons

sghiaa:
[Add "toggle on all" and "toggle off all"](https://app.asana.com/0/584764604969369/705930618035020)

### DIFF
--- a/assets/css/signs.css
+++ b/assets/css/signs.css
@@ -31,6 +31,10 @@
   padding-bottom: 6px;
 }
 
+.btn {
+  border: 2px solid #386BB2;
+}
+
 input[type="checkbox"].ios8-switch {
   position: absolute;
   margin: 8px 0 0 16px;
--- a/assets/css/signs.css
+++ b/assets/css/signs.css
@@ -31,7 +31,7 @@
   padding-bottom: 6px;
 }
 
-.btn {
+.toggle-btn {
   border: 2px solid #386BB2;
 }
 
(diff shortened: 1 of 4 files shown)


assets/css/signs.css
```diff
@@ -31,6 +31,10 @@
   padding-bottom: 6px;
 }
 
+.btn {
```
kohlerjp: Maybe we should add a separate class instead of altering an existing bootstrap one?

sghiaa: i took inspiration from dotcom, which edits btn - but ill just rename it

assets/js/toggle-all.js
```diff
@@ -0,0 +1,15 @@
+document.querySelector("#toggle-on").addEventListener("click", function(event) {
+  var checkboxes = document.querySelectorAll('input[type="checkbox"]');
```
losvedir: Rather than selecting _all_ checkboxes on the page (which seems like it could inadvertently target some sort of header control or route switcher or something we add in the future), it would be good to scope it to just the checkboxes we care about. I think we should wrap the `for` loop that iterates and creates the toggles in a class `js-sign-toggles` (I think dotcom used the convention that HTML elements controlled via JS had a `js-` prefix, so we wouldn't accidentally change it), and then update the selector here to be `'.js-sign-toggles input[type="checkbox"]'`.

sghiaa: fair enough - pr to follow

PR #30, opened 2018-07-19
add endpoints to post messages and also to show the most recent message

sghiaa:
adds `/messages` which displays the most recent message weve received on `/messages/update`.  pr to come on realtime_signs which will add signs ui as an additional destination for our signs messages

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -21,6 +21,9 @@ config :signs_ui,
 # HTTP config
 config :signs_ui, :redirect_http?, false
 
+config :signs_ui,
+  api_key: "dXNlcm6hbWU6cGFzc3dvcmQ="
+
 # Configures Elixir's Logger
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
--- a/config/config.exs
+++ b/config/config.exs
@@ -22,7 +22,7 @@ config :signs_ui,
 config :signs_ui, :redirect_http?, false
 
 config :signs_ui,
-  api_key: "dXNlcm6hbWU6cGFzc3dvcmQ="
+  api_key: "placeholder_key"
 
 # Configures Elixir's Logger
 config :logger, :console,
--- a/lib/signs_ui/signs/messages.ex
+++ b/lib/signs_ui/signs/messages.ex
@@ -1,0 +1,29 @@
+defmodule SignsUi.Signs.Messages do
+  use GenServer
+
+  @type message :: String.t
+
+  def start_link(opts \\ []) do
+    GenServer.start_link(__MODULE__, opts, opts)
+  end
+
+  def init(_opts) do
+    {:ok, []}
+  end
+
+  def list_messages(pid \\ __MODULE__) do
+    GenServer.call(pid, :list_messages)
+  end
+
+  def add_message(pid \\ __MODULE__, message) do
+    GenServer.call(pid, {:add_message, message})
+  end
+
+  def handle_call(:list_messages, _from, messages) do
+    {:reply, messages, messages}
+  end
+
+  def handle_call({:add_message, message}, _from, messages) do
+    {:reply, {:ok, [message]}, [message]}
+  end
+end
--- a/lib/signs_ui/signs/messages.ex
+++ b/lib/signs_ui/signs/messages.ex
@@ -24,6 +24,7 @@ defmodule SignsUi.Signs.Messages do
   end
 
   def handle_call({:add_message, message}, _from, messages) do
-    {:reply, {:ok, [message]}, [message]}
+    messages = Enum.take([message | messages], 10)
+    {:reply, {:ok, messages}, messages}
   end
 end
--- a/lib/signs_ui_web/controllers/messages_controller.ex
+++ b/lib/signs_ui_web/controllers/messages_controller.ex
@@ -5,15 +5,15 @@ defmodule SignsUiWeb.MessagesController do
 
   def index(conn, _params) do
     messages = Messages.list_messages()
-    render(conn, "index.html", messages: messages, token: get_csrf_token())
+    render(conn, "index.html", messages: messages)
   end
 
   def create(conn, params) do
     case Messages.add_message(params) do
       {:ok, messages} ->
-        render(conn, :index, messages: messages, token: get_csrf_token())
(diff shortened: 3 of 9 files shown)


lib/signs_ui_web/router.ex
```diff
@@ -29,9 +29,12 @@ defmodule SignsUiWeb.Router do
     get "/", PageController, :index
     get "/signs", SignsController, :index
     post "/signs", SignsController, :update
+    get "/messages", MessagesController, :index
   end
 
   scope "/", SignsUiWeb do
+    pipe_through [:api]
+    post "/messages/update", MessagesController, :update
```
losvedir: The conventional endpoint here would probably be `post "/messages`, MessageController, :create`

losvedir: Also, I don't think we should mess with the `/_health` endpoint. It doesn't accept JSON, for example. You should probably put this in its own scope.

sghiaa: thats true, originally i had this as updating a list of messages, so i should probably make this change to reflect what it really is now.

lib/signs_ui/signs/messages.ex
```diff
@@ -0,0 +1,29 @@
...
+  def add_message(pid \\ __MODULE__, message) do
+    GenServer.call(pid, {:add_message, message})
+  end
+
+  def handle_call(:list_messages, _from, messages) do
+    {:reply, messages, messages}
+  end
+
+  def handle_call({:add_message, message}, _from, messages) do
+    {:reply, {:ok, [message]}, [message]}
```
losvedir: Actually, as a proof of concept, might as well make it hold several messages (I mean, looking at the view it _seems_ to support rendering multiple messages, right?).

I think

```ex
messages = Enum.take([message | messages], 10)
{:reply, {:ok, messages}, messages}
```

is all you have to do... when you were mentioning in slack about just doing the simplest approach, I thought to avoid iterating through them in the view or whatever, but if you've already done that...

lib/signs_ui_web/router.ex
```diff
@@ -29,9 +29,12 @@ defmodule SignsUiWeb.Router do
     get "/", PageController, :index
     get "/signs", SignsController, :index
     post "/signs", SignsController, :update
+    get "/messages", MessagesController, :index
   end
 
   scope "/", SignsUiWeb do
+    pipe_through [:api]
+    post "/messages/create", MessagesController, :create
```
losvedir: Specifically, from my last comment, it would be `post /messages`. The HTTP method POST is itself enough to imply that you're creating a new resource, no need to add it to the URL as well.

lib/signs_ui/signs/messages.ex
sghiaa: yeah, the view is probably a relic of when i was displaying all of the messages ive recieved.  i agree, might as well show more than 1 then.

lib/signs_ui_web/router.ex
losvedir: That's what `resources "/messages", MessagesController, only: [:create]` would do for you here. Since we're only making one route here, I don't think we need to use that macro, but we might as well follow the conventions that it establishes.

lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -0,0 +1,19 @@
+defmodule SignsUiWeb.MessagesController do
+  use SignsUiWeb, :controller
+
+  alias SignsUi.Signs.Messages
+
+  def index(conn, _params) do
+    messages = Messages.list_messages()
+    render(conn, "index.html", messages: messages, token: get_csrf_token())
```
losvedir: As I mentioned in slack, I don't think we need to deal with this csrf token stuff, but we do need some sort of authentication. I think a hard coded key we pass as a header in the request should be fine.

sghiaa: oh, right - i knew i forgot to remove something.  currently the index requires the basic auth stuff - i can add some key to the create side of it though

config/config.exs
```diff
@@ -21,6 +21,9 @@ config :signs_ui,
 # HTTP config
 config :signs_ui, :redirect_http?, false
 
+config :signs_ui,
+  api_key: "dXNlcm6hbWU6cGFzc3dvcmQ="
```
losvedir: Can you make this `"API_KEY"` or something more obviously just a place holder?

paulswartz: If this is a real API key, it should be rebased off the branch as well.

lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -0,0 +1,19 @@
...
+    messages = Messages.list_messages()
+    render(conn, "index.html", messages: messages)
+  end
+
+  def create(conn, params) do
+    case Messages.add_message(params) do
+      {:ok, messages} ->
+        send_resp(conn, 201, "")
+      {:error, _} ->
+        render(conn, "index.html", messages: [])
```
losvedir: I think here we also want just a short `send_resp(conn, 400, "")`.

test/signs_ui_web/controllers/messages_controller_test.exs
```diff
@@ -0,0 +1,35 @@
...
+    test "lists all messages", %{conn: conn} do
+      conn = conn
+             |> add_req_header(:basic)
+             |> get(messages_path(conn, :index))
+      assert html_response(conn, 200) =~ "Listing Messages"
+    end
+  end
+
+  describe "create messages" do
+    test "responds with the json message when its valid", %{conn: conn} do
```
losvedir: Can you also have one where the API key isn't valid?

config/config.exs
sghiaa: yeah

lib/signs_ui_web/controllers/messages_controller.ex
sghiaa: thats a fair point

test/signs_ui_web/controllers/messages_controller_test.exs
sghiaa: i can

PR #32, opened 2018-07-19
Webpack and react and channels, oh my!

losvedir:
This PR has two commits:

1) First we change the frontend tooling from brunch to webpack, which is the general standard for react.

2) Adds react, and a sample app with a genserver that loops every couple seconds and broadcasts a message over the channel. If you have `/test` open, then it will append that message to the list.

The first commit needs to go in regardless. The second one is a proof of concept that can be tweaked to start working towards the actual countdown viewer. The PR could merge now, and Alex's could merge, and then we'd issue a new PR to unify. Or I could wait until his goes in, and just unify here.


paulswartz:
I'm not opposed to JS, but is there a reason this can't be a normal page that reloads periodically?

losvedir:
I wouldn't say it "can't" be done as a normal page that reloads periodically, but I think the experience for the PIOs will be nicer if it's kept up to date in real time, and better for us to debug the signs app, as well as being a little more forward looking.

We and the PIOs will probably mostly be using this to respond to bug reports and to get at-a-glance understanding of the system, probably by seeing all of the signs at a time, or maybe all the signs on a line. We update the red line signs about 150 times / minute, and if one of those updates is glitchy we might miss it. We've had an issue in the past where we had the expiration wrong and had lines on the sign go blank for a second or two, and another issue where the line briefly fluttered "5 min" -> "6 min" -> "5 min" so it would be handy to see those happening in real time as well.

We went with the simple refresh every couple seconds approach with our dashboard viewer, and while it works, we don't love it. Issues include stuttering/repainting during the reload and occasionally losing scroll position.

Lastly, our plan is to eventually integrate the viewer with the signs control toggles (and add an ability to set static messages). Automatic refreshes could interfere with that, if a refresh happened while a PIO was in the midst of making a change.

I think the downsides to a JS heavy approach are:

1) Compatibility issues - but it's only for us and the PIOs and we all use modern browsers
2) Testing - I didn't add any tests in this PR yet, but we had a pretty good react-friendly JS testing harness at my last company that I plan to introduce here, once we actually have stuff to test.
3) Complexity - I actually love react and I think it simplifies day-to-day development, but it does add a lot of complexity _around_ the development related to building and deploying and the JS ecosystem. In my experience yarn with its lock file helps things in that regard some (I could never get npm's lock file to work quite right at my last company), and react is fairly stable at this point. There will probably be more time towards upkeep than with a vanilla Phoenix app and no JS on the frontend, but I wouldn't expect _tons_ more.

Are there other aspects I'm not thinking of? I actually don't mind frontend development (when I can target modern browsers and use react), so I'm not discounting it so much when making trade offs vs. phoenix development. A slight improvement in user experience and a foundation for a more interactive app in the future are enough to warrant the use of JS to me, but if we have an organization policy to avoid JS unless strongly warranted then it could still be made to work moderately well without it.

paulswartz:
If all you're trying to do is avoid page reloads, there are simpler ways to do that without introducing a new framework.

Did you talk with Sky about the Glides experience with Elm?

gshkolnik:
In the same vein, do we really need to add another framework to our stack?

losvedir:
I'll check with Sky, though. I'm not opposed to standardizing on Elm if that's what we want... is it? None of us on RTR know it but it probably wouldn't be _too_ hard to get up to speed. But if we do want to make our org-wide stack Elixir/Elm, it should probably be a larger discussion.

paulswartz:
Having a larger discussion sounds like a good idea: we had a discussion when Glides went w/ Elm already. I do think that the general rule should be to err on the side of using tools that other teams are already using, rather than bringing in new ones. For example, I wouldn't be :100: opposed to using Rust for a project (seems like an interesting language!) but there would need to be a reason why one of our existing languages wouldn't be a good choice.

losvedir:
Fair enough. I didn't realize there was already a larger discussion around when Glides chose it. I chatted with Sky and it sounds like it's working really well for Glides. @dmaltzan, @sghiaa, and I are interested in giving elm a shot, so I'll take a stab at redoing this PR with elm instead of react.

ryan-mahoney:
@gshkolnik @paulswartz when are we having this JS discussion?  Back when Elm was selected for GLIDES, I think the discussion was framed around which typed JS flavor should we use, not which Framework. In the case of Elm, it happens to also be a language and framework... but I think we should be open to other options.

PR #33, opened 2018-07-23
move api key out of config

sghiaa:


lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -10,7 +10,7 @@ defmodule SignsUiWeb.MessagesController do
 
   def create(conn, params) do
     case Messages.add_message(params) do
-      {:ok, messages} ->
+      {:ok, _messages} ->
```
dmaltzan: what's this part?

sghiaa: there was a compile warning so  i guess i drive by'd it.  i can do it separately if youd prefer.

dmaltzan: nope that's fine.

PR #34, opened 2018-07-24
And here's the frontend approach with Elm!

losvedir:


losvedir:
Closing in favor of the react approach.

PR #35, opened 2018-07-26
ASH/GJD Sign UI

losvedir:
Asana: [M: an admin can understand the messages in the form of a sign](https://app.asana.com/0/584764604969369/719454568907232)

screen shot 2018-07-26 at 9 39 18 am

There's a known bug where if realtime_signs tries to update both signs at the same time, only the bottom line is updated here. That will be addressed in a separate ticket.

Following a conversation with Ryan, and the conventions used in the `create-react-app` helper, I introduced Jest for JS testing and Enzyme for react testing. With Enzyme you can mount your JSX components and then simulate interacting with them (sending in clicks and such), and then querying and asserting against how it should behave. It's kind of in the middle of unit and integration tests. Slower and more in depth than a unit test, but much lighter than bringing in wallaby or something like that. Unit tests are of course possible as well. This PR adds some unit tests and an enzyme test. I'll update our semaphore to run the tests as part of CI.

### DIFF
--- a/assets/js/Sign.js
+++ b/assets/js/Sign.js
@@ -3,10 +3,12 @@ import React, { Component } from 'react';
 class Sign extends Component {
   render() {
     return (
-      <div className="sign">
-        <div>id: {this.props.signId}</div>
-        <div>lineOne: {this.props.lineOne}</div>
-        <div>lineTwo: {this.props.lineTwo}</div>
+      <div className="viewer--sign">
+        <div className="viewer--sign-id">{this.props.signId}</div>
+        <div className="viewer--sign-lines">
+          <div className="viewer--sign-line">{this.props.lineOne}</div>
+          <div className="viewer--sign-line">{this.props.lineTwo}</div>
+        </div>
       </div>
     );
   }
--- a/assets/js/Sign.js
+++ b/assets/js/Sign.js
@@ -1,0 +1,14 @@
+import React, { Component } from 'react';
+
+class Sign extends Component {
+  render() {
+    return (
+      <div className="sign">
+        <div>id: {this.props.signId}</div>
+        <div>content: {JSON.stringify(this.props.content)}</div>
+      </div>
+    );
+  }
+}
+
+export default Sign;
--- a/assets/js/Sign.js
+++ b/assets/js/Sign.js
@@ -5,7 +5,8 @@ class Sign extends Component {
     return (
       <div className="sign">
         <div>id: {this.props.signId}</div>
-        <div>content: {JSON.stringify(this.props.content)}</div>
+        <div>lineOne: {this.props.lineOne}</div>
+        <div>lineTwo: {this.props.lineTwo}</div>
       </div>
     );
   }
--- a/assets/js/Viewer.js
+++ b/assets/js/Viewer.js
@@ -6,10 +6,10 @@ class Viewer extends Component {
     return(
       <ul>
         {Object.keys(this.props.signs).map( (key, index) => {
-
+          let lines = this.props.signs[key];
+          return <li key={key}><Sign signId={key} lineOne={lines[0]} lineTwo={lines[1]}></Sign></li>
         })}
       </ul>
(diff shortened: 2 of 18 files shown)


assets/js/Viewer.js
```diff
@@ -1,17 +1,21 @@
...
   render() {
     return(
-      <ul>
-        {this.props.messages.map( msg => {
-          return(
-            <li key={msg}>{msg}</li>
-          );
+      <div>
+        {this.sortedKeys(this.props.signs).map( (key, index) => {
+          let lines = this.props.signs[key];
```
meagonqz: let/const are blocked scoped so this can be const since you don't change it

I'd recommend adding eslint with the airbnb config to catch stuff like this, it's a good ES6 teacher. https://www.npmjs.com/package/eslint-config-airbnb

assets/js/ViewerApp.js
```diff
@@ -19,17 +21,17 @@ class ViewerApp extends Component {
       .join()
       .receive("ok", response => {console.log("Joined successfully", response)});
 
-    channel.on("new_msg", payload => {
+    channel.on("sign_update", ({sign_id, line_number, text}) => {
       this.setState({
```
meagonqz: If you're using setState with this.state, due to the batching of setState, you want to pass a function to setState, not an object.

https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

assets/js/Viewer.js
```diff
@@ -1,17 +1,21 @@
...
 class Viewer extends Component {
   render() {
     return(
-      <ul>
-        {this.props.messages.map( msg => {
-          return(
-            <li key={msg}>{msg}</li>
-          );
+      <div>
+        {this.sortedKeys(this.props.signs).map( (key, index) => {
```
meagonqz: It doesn't look like you use index so you can just remove it.

assets/js/Sign.js
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
+
+class Sign extends Component {
+  render() {
```
meagonqz: it's common to do something like `const { signId, lineOne, lineTwo } = this.props` at the top of a render declaration, and then you could just do {signId} below...stylistic choice though

lib/signs_ui/signs/messages.ex
```diff
@@ -24,8 +24,22 @@ defmodule SignsUi.Signs.Messages do
   end
 
   def handle_call({:add_message, message}, _from, messages) do
-    SignsUiWeb.Endpoint.broadcast!("signs:all", "new_msg", %{body: "#{inspect(message)}"})
-    messages = Enum.take([message | messages], 10)
+    sta = message["sta"]
+    [duration, zone_line, text] = String.split(message["c"], ["~", "-"])
```
losvedir: I have some comments on parsing this info generally (e.g. maybe using a Regex, and allowing for the different components to be in a different order, which I think is allowed), but it's probably all superseded by the custom parser stuff you're looking into now related to Phoenix eating the duplicate command, so I think this is fine for our first pass.

assets/js/Viewer.js
```diff
@@ -1,17 +1,21 @@
...
+          let lines = this.props.signs[key];
+          return <Sign key={key} signId={key} lineOne={lines[0]} lineTwo={lines[1]}></Sign>
         })}
-      </ul>
+      </div>
     );
   }
+
+  sortedKeys(signs) {
+    return Object.keys(signs).slice(0).sort();
```
meagonqz: I don't think you need the `.slice(0)` here

losvedir:
@meagonqz - Awesome, thanks so much for the review and pointers! I added `eslint` with the `airbnb` config and updated CI to require it to pass. It was a pretty good learning experience making all the changes it required.

assets/js/Viewer.js
losvedir: Ah, yeah. Good catch.

PR #36, opened 2018-07-26
parse body so that we can see duplicate c's in the post

sghiaa:


lib/signs_ui_web/realtime_signs_body_reader.ex
```diff
@@ -0,0 +1,21 @@
+defmodule RealtimeSignsBodyReader do
+  def read_body(conn, opts) do
+    {:ok, body, conn} = Plug.Conn.read_body(conn, opts)
+    conn = update_in(conn.assigns[:raw_body], &[body | &1 || []])
```
losvedir: Like the example in the docs you took this from, I think it's probably best to just return `{:ok, body, conn}` here. Not every request will have to deal with duplicate params, and so we don't really want to have to decode, change, re-encode, and then decode _again_ later, if we don't have to. I think the point of this parser is to just stash the body so we have access to it later. So in the controller, you'll be able to access `:raw_body` in the conn, and can do any manipulation you want there.

sghiaa: wouldnt that mean we have to look at the raw body to make sure we got everything in the controller?  to me it seems like we would want to have the body be the body intended by the request.

losvedir: Alternatively, since I guess it will get decoded again either way, maybe you _don't_ need to store the `raw_body` here, and we'll just rely on the correctly parsed data in the controller. Basically we either _fix_ the parsing here, or we stash the raw body in `:raw_body` to parse later, but I don't think we need to do both.

I'm also a little curious if you can use this, instead of hardcoding in `c` and `c2`, to store the map key as `c[]` and the value as a list. I think then in the controller when you access the `"c"` param will actually get a list of all the values.

sghiaa: i tried doing it as a list initially, but it was failing to encode correctly.  i guess urlencoding doesnt like lists?

paulswartz:
What was the encoded body with the list?

sghiaa:
@paulswartz i think it ended up being that i wasnt encoding the right way.  it seems to be working with `Plug.Conn.Query.encode`

PR #37, opened 2018-08-02
send expiration time to react and then dont display the line if its e…

sghiaa:
[M: handle expiration time in countdown viewer](https://app.asana.com/0/584764604969369/753278615983449)

the idea with this app is we want to as closely replicate what is on actual signs as possible.  to do this, we need to also replicate how the signs expire messages after a given expiration time.  As of right now, our realtime_signs app will only ever pass expiration times as the argument `e###`, meaning expires after ### seconds.  technically it is possible to pass `t###`, which will be "expires at time ###" but we never use that so it is unimplemented here.

signs-dev.mbtace.com/messages
pm me for username/password

### DIFF
--- a/assets/js/Sign.jsx
+++ b/assets/js/Sign.jsx
@@ -1,20 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
+function displayLine(duration, currentTime) {
+  return (Date.parse(duration) - currentTime) > 0;
+}
+
 function Sign({
   signId, lineOne, lineOneDuration, lineTwo, lineTwoDuration, currentTime,
 }) {
-  const remainingTimeOne = Date.parse(lineOneDuration) - currentTime;
-  const remainingTimeTwo = Date.parse(lineTwoDuration) - currentTime;
-  let topLine = null;
-  let bottomLine = null;
-  if (remainingTimeOne > 0) {
-    topLine = lineOne;
-  }
-  if (remainingTimeTwo > 0) {
-    bottomLine = lineTwo;
-  }
-
   return (
     <div className="viewer--sign">
       <div className="viewer--sign-id">
@@ -22,10 +15,10 @@ function Sign({
       </div>
       <div className="viewer--sign-lines">
         <div className="viewer--sign-line">
-          {topLine}
+          { displayLine(lineOneDuration, currentTime) ? lineOne : null }
         </div>
         <div className="viewer--sign-line">
-          {bottomLine}
+          { displayLine(lineTwoDuration, currentTime) ? lineTwo : null }
         </div>
       </div>
     </div>
--- a/assets/js/Sign.jsx
+++ b/assets/js/Sign.jsx
@@ -41,4 +41,11 @@ Sign.propTypes = {
   currentTime: PropTypes.number.isRequired,
 };
 
+Sign.defaultProps = {
+  lineOne: null,
+  lineOneDuration: '0',
+  lineTwo: null,
+  lineTwoDuration: '0',
+};
+
 export default Sign;
--- a/assets/js/Sign.jsx
+++ b/assets/js/Sign.jsx
@@ -1,7 +1,18 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-function Sign({ signId, lineOne, lineTwo }) {
+function Sign({ signId, lineOne, lineOneDuration, lineTwo, lineTwoDuration, currentTime }) {
+  let remainingTimeOne = Date.parse(lineOneDuration) - currentTime
+  let remainingTimeTwo = Date.parse(lineTwoDuration) - currentTime
+  let topLine = null
+  let bottomLine = null
+  if(remainingTimeOne > 0) {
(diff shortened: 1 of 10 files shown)


assets/js/Viewer.jsx
```diff
@@ -18,16 +18,28 @@ class Viewer extends Component {
       <div>
         {this.sortedKeys().map((key) => {
```
meagonqz: I wouldn't bother with this class method `sortedKeys`. I'd just add the `const { signs } = this.props` at the top of the render method and call `Object.keys(signs).sort().map(key => ...` here, or `const sortedKeys = Object.keys(signs).sort()` and `sortedKeys.map(key => ...`

assets/js/Viewer.jsx
```diff
@@ -18,16 +18,28 @@ class Viewer extends Component {
...
+          );
         })}
       </div>
     );
   }
 }
 
 Viewer.propTypes = {
-  signs: PropTypes.objectOf(PropTypes.arrayOf(PropTypes.string)).isRequired,
+  signs: PropTypes.objectOf(PropTypes.arrayOf(PropTypes.object)).isRequired,
```
meagonqz: I'd use `PropTypes.shape({ attr: PropTypes.string, ... })` here instead of just PropTypes.object

assets/js/ViewerApp.jsx
```diff
@@ -22,26 +24,38 @@ class ViewerApp extends Component {
...
+    }) => {
       this.setState(prevState => ({
-        signs: updateSigns(prevState.signs, { signId, lineNumber, content }),
+        signs: updateSigns(prevState.signs, {
+          signId, duration, lineNumber, content,
+        }),
       }));
     });
+
+    window.setInterval(this.updateTime, 1000);
```
meagonqz: You should probably clear this interval with `clearInterval(id)` on unmount (on `componentWillUnmount`) see a similar example here: https://reactjs.org/docs/state-and-lifecycle.html#adding-lifecycle-methods-to-a-class

assets/js/ViewerApp.jsx
```diff
@@ -22,26 +24,38 @@ class ViewerApp extends Component {
...
+  updateTime() {
+    this.setState(({
+      currentTime: Date.now(),
+    }));
   }
 
   render() {
     const { signs } = this.state;
-
+    const { currentTime } = this.state;
```
meagonqz: You can combine this with the above to be `const { signs, currentTime } = this.state;`

assets/js/Viewer.jsx
```diff
@@ -18,16 +18,28 @@ class Viewer extends Component {
       <div>
         {this.sortedKeys().map((key) => {
           const { signs } = this.props;
+          const { currentTime } = this.props;
```
meagonqz: Can combine these to be `const { signs, currentTime } = this.props`

assets/js/Viewer.jsx
sghiaa: oh, thats good to know about, thanks!

assets/js/ViewerApp.jsx
sghiaa: cool, thanks!

assets/js/Viewer.jsx
sghiaa: sounds good to me

sghiaa: i ended up getting linter complaints about this not being stateless, so i will leave it as is for now since its not really being touched by this pr anyway.

meagonqz: I was baffled by this so I changed it locally and it passes lint, so I'm just going to push the baby commit with this change

assets/js/Sign.jsx
```diff
@@ -1,18 +1,31 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-function Sign({ signId, lineOne, lineTwo }) {
+function Sign({
+  signId, lineOne, lineOneDuration, lineTwo, lineTwoDuration, currentTime,
+}) {
+  const remainingTimeOne = Date.parse(lineOneDuration) - currentTime;
```
meagonqz: you could extract this into a separate function in this file, that takes duration and currentTime and returns a boolean, true if it's greater than 0, false otherwise.

and then in the return JSX:

{ remainingTimeOne ? lineOne : null }
...
{ remainingTimeTwo ? lineTwo : null }

sghiaa: good idea

assets/js/ViewerApp.jsx
```diff
@@ -22,26 +24,46 @@ class ViewerApp extends Component {
...
   }
 
-  render() {
-    const { signs } = this.state;
+  componentWillUnmount() {
+    clearInterval(this.timerID);
+  }
 
+  updateTime() {
+    this.setState(({
```
meagonqz: nitpick: I think this gained an extra set of () when it was previously a function (which you don't need to pass it as in this case)

sghiaa: whoops

PR #39, opened 2018-08-06
quick fix for the crash if we get a multipart message

sghiaa:
we dont want to crash if the message we receive does has more than 3 parts in anticipation of the paginated messages we want to handle in the near future.

### DIFF
--- a/lib/signs_ui/signs/messages.ex
+++ b/lib/signs_ui/signs/messages.ex
@@ -63,19 +63,24 @@ defmodule SignsUi.Signs.Messages do
   defp parse_commands(commands) do
     commands
     |> Enum.map(fn command ->
-      [duration, zone_line, text] = String.split(command, ["~", "-"])
+      [duration, zone_line_text] = String.split(command, ["~"])
       duration_regex = ~r/([a-z])([\d]+)/
       [_original, _expires?, duration] = Regex.run(duration_regex, duration)
 
-      text =
-        text
-        |> URI.decode()
-        |> String.trim("\"")
-        |> String.replace("+", " ")
+      [zone_line | text_list] = String.split(zone_line_text, ["-"])
+
+      text_list =
+        text_list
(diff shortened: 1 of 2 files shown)


test/signs_ui/signs/messages_test.exs
```diff
@@ -14,4 +14,18 @@ defmodule SignsUi.Signs.MessagesTest do
...
+    {:ok, pid} = SignsUi.Signs.Messages.start_link()
+
+    SignsUi.Signs.Messages.add_message(pid, %{
+      "MsgType" => "SignContent",
+      "c" => ["e130~n2-\"Alewife 12 min\".4-\"Braintre 4 min\""],
+      "sta" => "RDTC",
+      "uid" => "616722"
+    })
+
+    assert %{"RDTC-n" => [%{duration: _, text: ""}, %{duration: _, text: "Alewife 12 min\".4"}]} =
```
losvedir: Hmm, I don't know about the `\".4"` stuff. I guess it does fix the crash, which is useful, but maybe it should also only show whats in the quotes? I think this can be done pretty easily with a regex that captures what's between the quotes.

lib/signs_ui/signs/messages.ex
```diff
@@ -63,19 +63,31 @@ defmodule SignsUi.Signs.Messages do
...
-        text
-        |> URI.decode()
-        |> String.trim("\"")
-        |> String.replace("+", " ")
+      [zone_line | text_list] = String.split(zone_line_text, ["-"])
+
+      text_list =
+        text_list
+        |> Enum.map(fn text ->
+          text_time_regex = ~r/([\w\d\s]+).?([\d]?+)$/
```
losvedir: Right before the digit at the end, should that be a literal period `\.` maybe, rather than a wildcard?

sghiaa: oh, yeah -  good catch

PR #41, opened 2018-08-20
Ash line signs

sghiaa:
a quick pr to have branch tabs - not styled or anything, just strictly to be able to view by line.  style pr to follow if its a good enough priority.

### DIFF
--- a/assets/js/Line.jsx
+++ b/assets/js/Line.jsx
@@ -3,9 +3,6 @@ import PropTypes from 'prop-types';
 import Sign from './Sign';
 
 function lineSigns(line) {
-  if (line === '') {
-    return [];
-  }
   if (line === 'Mattapan') {
     return ['MCED-s', 'MCED-n',
       'MBUT-c',
@@ -84,12 +81,7 @@ function lineSigns(line) {
       'RBRA-c', 'RBRA-m',
       'RJFK-w', 'RJFK-s', 'RJFK-e', 'RJFK-n', 'RJFK-m'];
   }
-  return ['MCED-s', 'MCED-n',
-    'MBUT-c',
-    'MMIL-s', 'MMIL-n',
-    'MCEN-s', 'MCEN-n',
-    'MVAL-s', 'MVAL-n',
-    'MCAP-s', 'MCAP-n'];
+  return [];
 }
 
 function Line({ signs, currentTime, line }) {
--- a/assets/js/Line.jsx
+++ b/assets/js/Line.jsx
@@ -3,93 +3,93 @@ import PropTypes from 'prop-types';
 import Sign from './Sign';
 
 function lineSigns(line) {
-  if (line == "") {
-    return []
+  if (line === '') {
+    return [];
   }
-  if(line == "Mattapan") {
-    return ["MCED-s", "MCED-n",
-      "MBUT-c",
-      "MMIL-s", "MMIL-n",
-      "MCEN-s", "MCEN-n",
-      "MVAL-s", "MVAL-n",
-      "MCAP-s", "MCAP-n"]
+  if (line === 'Mattapan') {
+    return ['MCED-s', 'MCED-n',
+      'MBUT-c',
+      'MMIL-s', 'MMIL-n',
+      'MCEN-s', 'MCEN-n',
+      'MVAL-s', 'MVAL-n',
+      'MCAP-s', 'MCAP-n'];
   }
-  else if(line == "SL3") {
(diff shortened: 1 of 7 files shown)


assets/js/Line.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import Line from './Line';
+
+test('does not show messages that have expired', () => {
```
losvedir: This doesn't seem like the sort of thing that should be tested here. I think it's being tested by the Sign, right? Rather, this should test something about the correct signs appearing for a given line.

assets/js/Viewer.test.js
```diff
@@ -3,15 +3,14 @@ import { mount } from 'enzyme';
 
 import Viewer from './Viewer';
 
-test('does not show messages that have expired', () => {
-  const now = Date.now();
-  const expired = new Date(now).toLocaleString();
-  const fresh = new Date(now + 5000).toLocaleString();
-  const signs = { signID: [{ text: 'Alewife 1 min', duration: fresh }, { text: 'Alewife 3 min', duration: expired }] };
+test('Shows all signs for a line', () => {
```
losvedir: I see. You have the line showing/hiding test here. I think it makes more sense in `Line` since that's where the logic actually is.

However, I think something like this might make sense as a `ViewerApp` test, as something of an integration test. If you look at the [docs](https://airbnb.io/enzyme/docs/api/mount.html#full-rendering-api-mount) for our testing library, there's an example of `find(...).simulate('click')`. It would be nice to have a test of that, to exercise your `onClick` functions.

assets/js/ViewerApp.jsx
```diff
@@ -50,11 +57,56 @@ class ViewerApp extends Component {
     });
   }
 
+  blueLine() {
```
losvedir: Just FYI, in case you didn't know, there's a more concise way to do this:

```js
changeLine(line) {
  this.setState({ line });
}
```

and then use a lambda in the `onClick` to pass the argument:

```js
<button type="button" onClick={(e) => this.changeLine('Blue')}>
  Blue
</button>
```

However, lambdas in `render` are *usually* considered a bad practice. In fact, our linter might even complain. I know some linters do, but not sure about ours. (Maybe you had it this way originally and changed it to your current approach because of the linter?) The reasons it's considered are performance related, but for a component like this which shouldn't re-render frequently, and when it does will change the whole screen, it probably doesn't matter. So I'd suggest changing it here, to simplify the code, and have our (future) test coverage be easier unless our linter complains in which case it's okay to leave alone.

assets/js/Line.jsx
```diff
@@ -0,0 +1,143 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Sign from './Sign';
+
+function lineSigns(line) {
+  if (line === '') {
```
losvedir: This feels more like an `if` / `else if` chain to me than a series of `if`s. I know with `return` in each one it's effectively the same, but I think they convey a different intent. Also, rather than `'Mattapan'` being the fallback, how about making it `line === 'Mattapan'` and making this empty array be the final `else`?

assets/js/Viewer.test.js
sghiaa: yeah, ill add something like that and move the other tests around

assets/js/Line.jsx
sghiaa: the linter told me not to use else after a return, but i can move that to the end

assets/js/ViewerApp.jsx
sghiaa: good tip -  the linter likes it

assets/js/ViewerApp.test.js
```diff
@@ -0,0 +1,28 @@
...
+  };
+
+  const currentTime = now + 2000;
+  const line = 'Red';
+
+  const wrapper = mount(
+    React.createElement(ViewerApp, { initialSigns: signs, currentTime, line }, null),
+  );
+
+  wrapper.find('#red-button').simulate('click');
```
losvedir: I think this test is really cool.

assets/js/Viewer.test.js
losvedir: I feel like this test can be removed now since it's covered by the tests in Sign and Line. Is there any other functionality being tested here?

sghiaa: oh, i actually had removed it but i guess it didnt save - ill push that up

PR #42, opened 2018-08-27
Fix bug where we couldn't disable signs.

losvedir:
Asana: [Bug: prod Signs UI shows "forbidden" when you try to turn on/off a sign](https://app.asana.com/0/584764604969369/792472905559025)

From testing deploys on signs-ui-dev, I confirmed that it was #36 that introduced the issue. In that PR we pre-parse the body to prevent Phoenix from "eating" duplicate params, by putting them in the syntax that will later cause the `params` to be a list. However, we do that even for singular params, which was interfering with the CSRF token, preventing us from submitting forms.

I changed the pre-parsing to leave singular params alone, only put them in a list if there's multiple values for the same param.

### DIFF
--- a/lib/signs_ui/signs/messages.ex
+++ b/lib/signs_ui/signs/messages.ex
@@ -62,6 +62,7 @@ defmodule SignsUi.Signs.Messages do
 
   defp parse_commands(commands) do
     commands
+    |> List.wrap
     |> Enum.map(fn command ->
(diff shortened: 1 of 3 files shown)


losvedir:
GitHub appears to have issues right now and isn't noticing my pushes to this branch. (Their status page shows they're looking into the errors). Going to try closing this PR and opening a new one, to see if it has my newest commits.

PR #46, opened 2018-08-31
Beautify the frontend a bit

losvedir:
Asana: [L: Styling line-page viewer](https://app.asana.com/0/584764604969369/795862646793285)

Deploying the branch now to signs-dev so you can click around and see it. I'm working on tests now.


assets/css/viewer.css
```diff
@@ -1,22 +1,36 @@
...
+  background-color: rgb(206, 206, 206);
+  margin-bottom: 1em;
+  width: 20em;
 }
 
 .viewer--sign-id {
-  font-weight: bold;
-  padding: 0.5em;
+  border-top: 10px solid orange;
+  /*padding-top: 0.5em;*/
```
sghiaa: do we need these comments

assets/js/Station.jsx
```diff
@@ -0,0 +1,106 @@
...
+          {makeSign(config, 's', signs, currentTime, line)}
+          {makeSign(config, 'w', signs, currentTime, line)}
+        </div>
+        <div className="col">
+          {makeSign(config, 'c', signs, currentTime, line)}
+          {makeSign(config, 'm', signs, currentTime, line)}
+        </div>
+        <div className="col">
+          {makeSign(config, 'n', signs, currentTime, line)}
+          {makeSign(config, 'e', signs, currentTime, line)}
```
sghiaa: this causes weird results for example when on the orange line, e means busway and acts as a mezz - probably ok, but i thought id mention

assets/js/mbta.js
```diff
@@ -0,0 +1,532 @@
...
+      name: 'Oak Grove',
+      zones: {
+        n: true, s: true, e: 'Busway', w: false, c: false, m: true,
+      },
+    },
+    {
+      id: 'OMAL',
+      name: 'Malden Center',
+      zones: {
+        n: false, s: false, e: false, w: false, c: false, m: 'Entire Station',
```
sghiaa: maybe 'Both Platforms'?  or are there real m signs too

assets/js/Station.jsx
losvedir: Yeah, true. Maybe another reason (beyond OCHN/OCHS, etc) to change to a data structure of just specifying per-station, what signs we want where. Initially I was not a fan of putting presentational information in the configuration, but I'm coming around on not really seeing another way of doing it. But it's probably not worth doing now; rather just get this in and move on.

assets/js/mbta.js
losvedir: Not sure. I took the phrasing from the ID in realtime_signs: `malden_center_entire_station`. Does @dmaltzan know the layout of the station?

assets/css/viewer.css
```diff
@@ -1,23 +1,35 @@
...
-  width: 15em;
+  background-color: rgb(206, 206, 206);
+  margin-bottom: 1em;
+  width: 20em;
 }
 
 .viewer--sign-id {
-  font-weight: bold;
-  padding: 0.5em;
+  border-top: 10px solid orange;
```
sghiaa: just noticed that this is `orange` wheras other colors are given rgb values.  is orange what we wanted?

assets/js/mbta.js
dmaltzan: I think there are some on the mezzanine and some in the center of the platform

assets/css/viewer.css
dmaltzan: if you want to be really fancy the colors are available via the API as in `https://api-v3.mbta.com/routes/Orange`

dmaltzan: but that's probably overkill - you could just take it from that API call which has it as `ED8B00`

losvedir: The actual colors get set with JS, and can be found in `colors.js`. I got their values from the [MBTA color style guide](https://projects.invisionapp.com/dsm/mbta-customer-technology/digital-style-guide/folder/colors/5ac782001c780e0011d2f352), and I checked the orange against the one in the API here and it was the same, so I assume the rest are as well.

I think a border requires a width, style, and color, so I just used a default color here in order to set the width and style, so that the JS can later change it to route specific values. The color here should never actually display.

I'll just change this to `white`, then.

PR #47, opened 2018-09-05
show the stops when stopped away

sghiaa:


lib/signs_ui/signs/messages.ex
```diff
@@ -97,7 +97,15 @@ defmodule SignsUi.Signs.Messages do
 
       {zone, line_no} = String.split_at(zone_line, 1)
       line_no = String.to_integer(line_no)
-      {duration, zone, line_no, List.first(text_list)}
+
+      display_text =
+        case text_list do
+          [_stopped, n_stops, _away] -> n_stops
```
losvedir: I think the message comes in as `[n_stops, _away, _stopped]`.

sghiaa: oh, ok.  is that because of sorting alphabetically? i  see the to_string at least had the order as stopped n_stops away

```
 63     def to_string(%{headsign: headsign, stops_away: n}) do
 64       stop_word = if n == 1, do: "stop", else: "stops"
 65
 66       pages = [
 67         Content.Utilities.width_padded_string(headsign, "Stopped", 18),
 68         Content.Utilities.width_padded_string(headsign, "#{n} #{stop_word}", 18),
 69         Content.Utilities.width_padded_string(headsign, "away", 18)
 70       ]
 71
 72       {pages, 3}
 73     end
```
which is what i went based on

sghiaa: oh wait the to string is the signs display so i guess thats obviously how this would be.

dmaltzan: Yeah it comes in out of order because of a bug in the actual ARINC code which we of course have to replicate lol

test/signs_ui/signs/messages_test.exs
```diff
@@ -54,6 +54,20 @@ defmodule SignsUi.Signs.MessagesTest do
...
+    {:ok, pid} = SignsUi.Signs.Messages.start_link()
+
+    SignsUi.Signs.Messages.add_message(pid, %{
+      "MsgType" => "SignContent",
+      "c" => ["e130~n2-\"Alewife stopped\".4-\"Alewife 4 stops\".4-\"Alewife away\".4"],
+      "sta" => "RDTC",
+      "uid" => "616722"
+    })
+
+    assert %{"RDTC-n" => [%{duration: _, text: ""}, %{duration: _, text: "Alewife 4 stops"}]} =
```
dmaltzan: Are we really only expecting one space between `Alewife` and `4 stops`?

sghiaa: no, but i dont think that matters for this test

dmaltzan: Oh I get it, duh.  Carry on

PR #48, opened 2018-09-07
Control signs from viewer

losvedir:
Asana: [L: Add sign controller buttons to line page viewer](https://app.asana.com/0/584764604969369/795862646793286).

This is on https://signs-dev.mbtace.com.

screen shot 2018-09-07 at 11 23 42 am

Still need to refactor a bit and add some tests, but opening the PR so Dave and Alex have somewhere to leave comments from playing around with it.


dmaltzan:
I'm not sure where this is coming from (maybe a config file that's not in source control?) but "Seaport Ave" is a wrong name on the UI - should be Eastern Ave

dmaltzan:
One more small bug report: "CP" is showing up in the smaller font size instead of the larger size of "NB" "SB" etc.

PR #62, opened 2019-01-15
Reads old/new format config, saves to new (backwards compatible) format

losvedir:
#### Summary of changes
**Asana Ticket:** A little piece of [[8] PIOs can put static text on a sign](https://app.asana.com/0/584764604969369/831665421652437)

This PR updates `signs_ui` to use the new config format we decided on. In particular, when the app is on this branch, it can read _either_ the old or the new format, but when it saves, it saves to the new format. (Note that the new format is backwards compatible, so `realtime_signs` will continue to work, even when we've saved with the new format.)

In addition, when it starts up, after reading in the JSON config file, it then immediately saves its state. This will update the format of all the signs.

Lastly, there's the legacy signs listing and control page (from before the react app). That will need to be removed eventually, since it doesn't support the different sign types, but for now I basically just prevented it from updating any state, rather than dealing with converting it to handle the new format.

#### Reviewer Checklist

### DIFF
--- a/lib/signs_ui/signs/request.ex
+++ b/lib/signs_ui/signs/request.ex
@@ -17,7 +17,7 @@ defmodule SignsUI.Signs.Request do
     case Poison.decode(json) do
       {:ok, response} ->
         {:ok,
-         Map.new(response, fn {sign_id, values} -> {sign_id, Sign.from_json(sign_id, values)} end)}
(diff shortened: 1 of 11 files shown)


losvedir:
Okay, I deployed to signs-dev. I toggled some signs there and they turned on/off correctly. I then restarted signs-dev (so it would re-read the JSON config file), and the same signs that were off were correctly read as off. I downloaded the config file from S3 and it looks like it's in the correct (new) format.

PR #63, opened 2019-01-17
Clean expired settings

lemald:
#### Summary of changes
**Asana Ticket:** [signs_ui automatically removes expired settings from JSON config](https://app.asana.com/0/0/989470430866018/f)

Added a new Genserver, also moved from using `Supervisor.Spec` to `Supervisor` for the overall application.

#### Reviewer Checklist
- [x] Meets ticket's acceptance criteria
- [x] Any new or changed functions have typespecs
- [x] Tests were added for any new functionality (don't just rely on Codecov)

### DIFF
--- a/lib/signs_ui/application.ex
+++ b/lib/signs_ui/application.ex
@@ -4,18 +4,24 @@ defmodule SignsUi.Application do
   alias SignsUi.Utilities.Config
 
   def start(_type, _args) do
-    import Supervisor.Spec
-
     set_runtime_config()
 
     # Define workers and child supervisors to be supervised
     children = [
       # Start the endpoint when the application starts
-      supervisor(SignsUiWeb.Endpoint, []),
-      worker(SignsUI.Signs.State, [[]]),
-      worker(SignsUi.Signs.Messages, [[name: SignsUi.Signs.Messages]])
-      # Start your own worker by calling: SignsUi.Worker.start_link(arg1, arg2, arg3)
-      # worker(SignsUi.Worker, [arg1, arg2, arg3]),
+      %{
+        id: Superviser,
+        start: {SignsUiWeb.Endpoint, :start_link, []},
+        type: :superisor
+      },
+      %{
+        id: State,
+        start: {SignsUI.Signs.State, :start_link, [[]]}
+      },
+      %{
+        id: Messages,
+        start: {SignsUi.Signs.Messages, :start_link, [[name: SignsUI.Signs.Messages]]}
+      }
     ]
 
     # See https://hexdocs.pm/elixir/Supervisor.html
--- a/lib/signs_ui/application.ex
+++ b/lib/signs_ui/application.ex
@@ -7,23 +7,10 @@ defmodule SignsUi.Application do
     set_runtime_config()
 
     children = [
-      %{
-        id: SignsUiWeb.Endpoint,
-        start: {SignsUiWeb.Endpoint, :start_link, []},
-        type: :supervisor
-      },
-      %{
-        id: SignsUI.Signs.State,
-        start: {SignsUI.Signs.State, :start_link, []}
-      },
-      %{
-        id: SignsUi.Signs.Messages,
-        start: {SignsUi.Signs.Messages, :start_link, [[name: SignsUi.Signs.Messages]]}
-      },
-      %{
-        id: SignsUI.Signs.Expiration,
-        start: {SignsUI.Signs.Expiration, :start_link, []}
-      }
+      SignsUiWeb.Endpoint,
+      SignsUI.Signs.State,
+      {SignsUi.Signs.Messages, [name: SignsUi.Signs.Messages]},
+      SignsUI.Signs.Expiration
     ]
 
     opts = [strategy: :one_for_one, name: SignsUi.Supervisor]
--- a/lib/signs_ui/application.ex
+++ b/lib/signs_ui/application.ex
@@ -18,7 +18,7 @@ defmodule SignsUi.Application do
       },
       %{
         id: SignsUi.Signs.Messages,
-        start: {SignsUi.Signs.Messages, :start_link, []}
+        start: {SignsUi.Signs.Messages, :start_link, [[name: SignsUi.Signs.Messages]]}
       }
     ]
 
--- a/lib/signs_ui/application.ex
+++ b/lib/signs_ui/application.ex
@@ -6,28 +6,24 @@ defmodule SignsUi.Application do
   def start(_type, _args) do
     set_runtime_config()
 
-    # Define workers and child supervisors to be supervised
     children = [
-      # Start the endpoint when the application starts
       %{
-        id: Superviser,
+        id: SignsUiWeb.Endpoint,
         start: {SignsUiWeb.Endpoint, :start_link, []},
         type: :supervisor
       },
       %{
-        id: State,
-        start: {SignsUI.Signs.State, :start_link, [[]]}
+        id: SignsUI.Signs.State,
+        start: {SignsUI.Signs.State, :start_link, []}
       },
       %{
-        id: Messages,
-        start: {SignsUi.Signs.Messages, :start_link, [[name: SignsUI.Signs.Messages]]}
+        id: SignsUi.Signs.Messages,
+        start: {SignsUi.Signs.Messages, :start_link, []}
       }
     ]
 
-    # See https://hexdocs.pm/elixir/Supervisor.html
-    # for other strategies and supported options
     opts = [strategy: :one_for_one, name: SignsUi.Supervisor]
-    Supervisor.start_link(children, opts)
+    {:ok, _pid} = Supervisor.start_link(children, opts)
   end
 
   # Tell Phoenix to update the endpoint configuration
--- a/lib/signs_ui/application.ex
+++ b/lib/signs_ui/application.ex
@@ -19,6 +19,10 @@ defmodule SignsUi.Application do
       %{
         id: SignsUi.Signs.Messages,
         start: {SignsUi.Signs.Messages, :start_link, [[name: SignsUi.Signs.Messages]]}
+      },
+      %{
+        id: SignsUI.Signs.Expiration,
+        start: {SignsUI.Signs.Expiration, :start_link, []}
       }
     ]
 
--- a/lib/signs_ui/application.ex
+++ b/lib/signs_ui/application.ex
@@ -12,7 +12,7 @@ defmodule SignsUi.Application do
       %{
         id: Superviser,
         start: {SignsUiWeb.Endpoint, :start_link, []},
-        type: :superisor
+        type: :supervisor
       },
       %{
         id: State,
--- a/lib/signs_ui/signs/expiration.ex
+++ b/lib/signs_ui/signs/expiration.ex
@@ -30,8 +30,6 @@ defmodule SignsUI.Signs.Expiration do
 
   @spec handle_info(:process_expired, state()) :: state()
   def handle_info(:process_expired, state) do
-    Logger.info("Running :process_expired loop")
-
     updates =
       state.sign_state_server
       |> SignsUI.Signs.State.get_all()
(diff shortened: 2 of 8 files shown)


lib/signs_ui/application.ex
```diff
@@ -4,24 +4,30 @@ defmodule SignsUi.Application do
...
 
-    # Define workers and child supervisors to be supervised
     children = [
-      # Start the endpoint when the application starts
-      supervisor(SignsUiWeb.Endpoint, []),
-      worker(SignsUI.Signs.State, [[]]),
-      worker(SignsUi.Signs.Messages, [[name: SignsUi.Signs.Messages]])
-      # Start your own worker by calling: SignsUi.Worker.start_link(arg1, arg2, arg3)
-      # worker(SignsUi.Worker, [arg1, arg2, arg3]),
+      %{
```
losvedir: You can use the simple shorthand for children:

```ex
children = [
  SignsUiWeb.Endpoint,
  SignsUI.Signs.State,
  {SignsUi.Signs.Messages, [[name: SignsUi.Signs.Messages]]},
  SignsUI.Signs.Expiration
]
```

You'll need to add `use GenServer` to your module, which it definitely should have, since it's a GenServer.

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,76 @@
...
+
+  @type state :: %{
+          time_fetcher: (() -> DateTime.t()),
+          loop_ms: integer(),
+          sign_state: pid()
+        }
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    name = opts[:name] || __MODULE__
```
losvedir: I know we haven't been very good about it elsewhere, but since `opts` is a `Keyword.t` it might be better here to work with it through the `Keyword` module, in particular `Keyword.get/3`.

lemald: Good point, I'll change it.

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,76 @@
+defmodule SignsUI.Signs.Expiration do
+  @moduledoc """
+  Handle cleaning up settings in the configuration that have since expired
+  """
+
+  @type state :: %{
+          time_fetcher: (() -> DateTime.t()),
+          loop_ms: integer(),
+          sign_state: pid()
```
losvedir: I find `sign_state` a confusing name here, since it implies to me the `%{id => sign}` map. In addition, it's not really a `pid` but a `GenServer.server()`. What about `sign_state_server: GenServer.server()` ?

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,76 @@
...
+    schedule_loop(self(), state.loop_ms)
+
+    {:noreply, state}
+  end
+
+  @spec expire_signs((() -> DateTime.t()), pid()) :: :ok
+  def expire_signs(time_fetcher, sign_state) do
+    signs =
+      SignsUI.Signs.State.get_all(sign_state)
+      |> Enum.map(&expire_single_sign(&1, time_fetcher))
```
losvedir: It might be nice to not update the state (POST to S3, broadcast to listening clients) if we don't have to. What about an approach something like:

```ex
expiration = time_fetcher.()

expired_signs = 
  sign_state_server
  |> SignsUI.Signs.State.get_all()
  |> Enum.flat_map(&expire_sign(&1, expiration)
  |> Enum.into()

if expired_signs != {} do
  SignsUI.State.update_some(sign_state_server, expired_signs)
end

:ok
```

And then `expire_sign/2` would have in it, roughly:

```ex
if expired? do
  [%Sign{...}]
else
  []
end
```

The "`flat_map` as a way to filter+map" approach is one we use fairly frequently here.

lib/signs_ui/application.ex
lemald: Done

lib/signs_ui/signs/expiration.ex
lemald: Done.

lib/signs_ui/signs/expiration.ex
lemald: Good thought, I'll go with that approach.

lemald: Done

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  @type state :: %{
+          time_fetcher: (() -> DateTime.t()),
+          loop_ms: integer(),
+          sign_state_server: GenServer.server()
+        }
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(opts \\ []) do
+    name = Keyword.get(opts, :name) || __MODULE__
```
losvedir: I think for all these it makes sense to make the default the optional third argument, rather than the `|| ...` construction.

lemald: Done

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,84 @@
...
+
+    if expired_signs != %{} do
+      SignsUI.Signs.State.update_some(sign_state_server, expired_signs)
+    else
+      :ok
+    end
+  end
+
+  @spec expire_single_sign(
+          {SignsUI.Signs.Sign.id(), SignsUI.Signs.Sign.t()},
```
losvedir: If you `alias SignsUI.Signs.Sign` in this file, and then refer to `Sign.id()`, `Sign.t()`, and `%Sign{..}` does it neaten this up a little bit? It's kind of hard for me to follow what's going on with all the line breaks and indents.

lemald: Oh, I had thought that to myself when I was working on it but never got around to actually doing it. I probably should.

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,82 @@
...
+    if expired_signs != %{} do
+      SignsUI.Signs.State.update_some(sign_state_server, expired_signs)
+    else
+      :ok
+    end
+  end
+
+  @spec expire_single_sign({Sign.id(), Sign.t()}, DateTime.t()) :: [{Sign.id(), Sign.t()}]
+  defp expire_single_sign(
+         {id, %Sign{config: %{expires: expiration}} = sign},
```
losvedir: I think you need a `when not is_nil(expiration)` since by the spec it's allowed to be `nil`, and I think `DateTime.compare(nil, _)` will :boom:.

lemald: I'll add that and also throw in an accompanying test.

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,82 @@
...
+         current_dt
+       ) do
+    if DateTime.compare(expiration, current_dt) == :lt do
+      [
```
losvedir: Just a small style preference, but if `mix format` allows this whole list element from line `62` to `68` on one line, could you do that?

lib/signs_ui/signs/expiration.ex
lemald: Done

lib/signs_ui/signs/sign.ex
```diff
@@ -59,7 +72,7 @@ defmodule SignsUI.Signs.Sign do
       id: sign_id,
       config: %{
         mode: :headway,
-        expires: config["expires"]
+        expires: expiration_from_string(config["expires"])
```
losvedir: By our JSON "spec" `config["expires"]` is allowed to be `nil` (meaning no expiration), so `expiration_from_string/1` should have a typespec of `String.t() | nil`, and another function clause that handles that.

lib/signs_ui/signs/sign.ex
```diff
@@ -147,6 +160,14 @@ defmodule SignsUI.Signs.Sign do
     }
   end
 
+  @spec expiration_to_iso8601(expires_on()) :: String.t() | nil
+  defp expiration_to_iso8601(expires) do
```
losvedir: Maybe two function clauses, both for brevity and for a bit of struct-assertion?

```ex
defp expiration_to_iso8601(%DateTime{} = dt), do: DateTime.to_iso8601(dt)
defp expiration_to_iso8601(nil), do: nil
```

lib/signs_ui/signs/sign.ex
lemald: Done

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,82 @@
...
+  def handle_info(:process_expired, state) do
+    expire_signs(state.time_fetcher, state.sign_state_server)
+
+    schedule_loop(self(), state.loop_ms)
+
+    {:noreply, state}
+  end
+
+  @spec expire_signs((() -> DateTime.t()), GenServer.server()) :: :ok
+  def expire_signs(time_fetcher, sign_state_server) do
```
losvedir: Actually, since this is the real "meat" of the server, what do you think about extracting out the core of it into a pure function, and move the global state interactions to the `handle_info` loop? That should make it easier to test a few more cases. That's the convention we try to shoot for in most of our code. (e.g. in `RTR`, the `RTR.Vehicles.State` module actually stores and retrieves the vehicles, and then the `RTR.Vehicles.Vehicles` module has a bunch of pure functions that take vehicles as an input and return different ones as the output.)

This function could receive the `get_all()` signs as an argument and then return its calculated `expired_signs`. The `get_all()` and the `if != {}` stuff would be in the `handle_info(:process_expired)` part.

In the `expiration_test.exs` file, you can then very easily pass in a few variety of signs in a map, some with `nil` and others with DateTimes in the past or future, etc, and assert that the correctly changed signs come out of the function.

The more end-to-end sort of test that you have now in `expiration_test.exs` could then be updated to be a test of `handle_info(:process_expired, ..)` instead, with minimal changes.

lib/signs_ui/signs/sign.ex
lemald: Done

lib/signs_ui/signs/expiration.ex
lemald: That sounds like an excellent idea for after I eat lunch. 🌯

lib/signs_ui/signs/sign.ex
```diff
@@ -49,6 +50,21 @@ defmodule SignsUI.Signs.Sign do
     %__MODULE__{id: id, config: %{mode: :off, expires: nil}}
   end
 
+  @spec expiration_from_string(String.t() | nil) :: expires_on()
+  defp expiration_from_string(expiration_string)
+       when not is_nil(expiration_string) and expiration_string != "" do
```
losvedir: I think this can be just `is_binary(expiration_string)`. A value of `""` is just as invalid as `"not-a-datetime"` and should trigger the same warning, IMO.

test/signs_ui/signs/expiration_test.exs
```diff
@@ -0,0 +1,74 @@
+defmodule SignsUI.Signs.ExpirationTest do
+  use ExUnit.Case
+  alias SignsUI.Signs.Sign
+
+  describe "expire_signs/2" do
+    test "expire_signs/2" do
```
losvedir: I like it. 👍

lib/signs_ui/signs/sign.ex
lemald: Done

test/signs_ui/signs/expiration_test.exs
lemald: Heh, ooops. I renamed the `test` to something a bit more descriptive

losvedir: Haha, I meant the refactored test, not just a sarcastic snark on the test name!

lib/signs_ui/signs/expiration.ex
```diff
@@ -0,0 +1,78 @@
...
+
+  @spec handle_info(:process_expired, state()) :: state()
+  def handle_info(:process_expired, state) do
+    updates =
+      state.sign_state_server
+      |> SignsUI.Signs.State.get_all()
+      |> expire_signs(state.time_fetcher)
+
+    if updates != %{} do
+      SignsUI.Signs.State.update_some(state.sign_state_server, updates)
```
losvedir: Maybe a `Logger.info` here, so we can more easily see if it's happening (or not)?

PR #66, opened 2019-01-23
Alert client when setting is expired

lemald:
#### Summary of changes
**No ticket, hotfix for:** [signs_ui automatically removes expired settings from JSON config](https://app.asana.com/0/0/989470430866018/f)

So that the UI updates when a sign expires

#### Reviewer Checklist

### DIFF
--- a/test/signs_ui/signs/state_test.exs
+++ b/test/signs_ui/signs/state_test.exs
@@ -6,7 +6,6 @@ defmodule SignsUI.Signs.StateTest do
   alias SignsUI.Signs.Sign
 
   @endpoint SignsUiWeb.Endpoint
-  @endpoint.subscribe("signs:all")
 
   describe "get_all/1" do
     test "Returns all signs" do
@@ -21,6 +20,7 @@ defmodule SignsUI.Signs.StateTest do
     test "updates some values and leaves others alone" do
       {:ok, pid} = GenServer.start_link(SignsUI.Signs.State, [], [])
 
+      @endpoint.subscribe("signs:all")
       token = Phoenix.Token.sign(SignsUiWeb.Endpoint, "user socket", "admin")
       {:ok, socket} = connect(SignsUiWeb.UserSocket, %{"token" => token})
       {:ok, _, _socket} = subscribe_and_join(socket, "signs:all", %{})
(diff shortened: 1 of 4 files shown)


losvedir:
This looks good, but can you add a test? Try following the docs [here](https://hexdocs.pm/phoenix/testing_channels.html) on testing channels.

lemald:
Yeah, I've been trying to add a test for the past hour. A lot of the channel testing seems different between Phoenix 1.3.2 and 1.4 and it turns out it's really hard to find the old Phoenix documentation online, so it's difficult going.

test/signs_ui/signs/state_test.exs
```diff
@@ -1,9 +1,13 @@
 defmodule SignsUI.Signs.StateTest do
   use ExUnit.Case
+  use Phoenix.ChannelTest
   import SignsUI.Signs.State
   alias SignsUI.Signs
   alias SignsUI.Signs.Sign
 
+  @endpoint SignsUiWeb.Endpoint
+  @endpoint.subscribe("signs:all")
```
losvedir: I think these should be part of the test, or part of a `setup` block if they need to happen before multiple. Changing state via module attributes is usually not so good, since they only get evaluated once, when the module is compiled. I guess that's not such a problem for this `.exs` file, but it still feels a little weird to me.

lemald: Well, I can't move it inside of the `test` because I get a `cannot set attribute @endpoint inside function/macro` error. I can put it inside the `describe`, though.

lemald: I can at least move the `@endpoint.subscribe` though if you think that's sufficient.

losvedir: You can't set a module attribute in a function, so I was thinking you'd just have a line `SignsUiWeb.Endpoint.subscribe("signs:all")` at the start of the test. But defining `@endpoint` outside and then doing `@endpoint.subscribe` in the test also works, I suppose.

lemald: Unfortunately if I get rid of `@endpoint` this causes issues for macros from `Phoenix.ChannelTest`, so it'll have to be the latter approach.

PR #81, opened 2019-05-08
Add basic routes for Cognito work

losvedir:
Adds a /cognito/new route to redirect to the AWS Cognito sign in page, and a /cognito route for it to redirect back to.

Not final work, just a baseline that we can each build off, to hopefully minimize conflicts.

### DIFF
--- a/lib/signs_ui_web/controllers/cognito_controller.ex
+++ b/lib/signs_ui_web/controllers/cognito_controller.ex
@@ -1,0 +1,25 @@
+defmodule SignsUiWeb.CognitoController do
+  use SignsUiWeb, :controller
+  require Logger
+
+  def index(conn, %{"code" => code}) do
+    Logger.info("cognito code=#{code}")
+
+    redirect(conn, to: "/")
+  end
+
+  def index(conn, _) do
+    Logger.info("cognito no_code")
+    send_resp(conn, 400, "no cognito code")
+  end
+
+  def new(conn, _) do
+    redirect(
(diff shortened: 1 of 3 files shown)


lib/signs_ui_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,25 @@
...
+  def index(conn, _) do
+    Logger.info("cognito no_code")
+    send_resp(conn, 400, "no cognito code")
+  end
+
+  def new(conn, _) do
+    redirect(
+      conn,
+      external:
+        "https://mbta-signs-dev.auth.us-east-1.amazoncognito.com/login?response_type=code&client_id=#{
```
lemald: Not sure if now is the time, but the `mbta-signs-dev` bit should probably be configurable on a per-environment basis for prod vs. dev.

lib/signs_ui_web/controllers/cognito_controller.ex
```diff
@@ -0,0 +1,25 @@
...
+          Application.get_env(:signs_ui, :cognito_client_id)
+        }&redirect_uri=http://localhost:4000/cognito"
```
lemald: Similarly I think the `redirect_uri` should be configurable.

losvedir:
Yeah, deliberate for now. At first I just had the whole string there for simplicity, but figured I should pull out the client ID since that's very slightly sensitive.

PR #86, opened 2019-05-14
Use Cognito for authentication

lemald:
#### Summary of changes
**Asana Ticket:** [[8] Use Amazon Cognito Single Sign-On for Signs UI](https://app.asana.com/0/584764604969369/1107845790526226/f)

Uses `ueberauth` to verify the user's login via Cognito and `guardian` to check authentication.

Basic changes are:
- Setting up our own Guardian stuff to generate and check tokens.
- Using the `ueberauth_cognito` strategy implemented [elsewhere](https://github.com/mbta/ueberauth_cognito).
- Updates to the router and controller to check for authentication and send the user through the workflow.
- Updates to the socket / channels to check for authentication and, if the user's login has expired, reloads the page rather than letting data get stale.

#### Reviewer Checklist


config/config.exs
```diff
@@ -20,11 +20,26 @@ config :signs_ui,
...
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   metadata: [:user_id]
 
+config :ueberauth, Ueberauth,
+  providers: [
+    cognito: {Ueberauth.Strategy.Cognito, []}
+  ]
+
+config :ueberauth, Ueberauth.Strategy.Cognito,
```
paulswartz: Maybe have this in `config/prod.exs` and a different one for local development?

lib/signs_ui_web/channels/signs_channel.ex
```diff
@@ -9,15 +9,53 @@ defmodule SignsUiWeb.SignsChannel do
...
+
+    case SignsUiWeb.AuthManager.decode_and_verify(token, claims) do
+      {:ok, _claims} -> true
+      {:error, _error} -> false
+    end
+  end
+
+  @spec maybe_send_auth_expired_message(Phoenix.Socket.t()) :: Phoenix.Socket.t()
+  defp maybe_send_auth_expired_message(socket) do
+    if socket.assigns[:auth_expired_message_sent] do
```
paulswartz: Since this is a boolean, something like `:auth_expired_message_sent?` (question mark) is appropriate.

lib/signs_ui_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,25 @@
+defmodule SignsUiWeb.AuthController do
+  use SignsUiWeb, :controller
+  plug(Ueberauth)
+
+  @spec callback(Plug.Conn.t(), map()) :: Plug.Conn.t()
+  def callback(%{assigns: %{ueberauth_auth: auth}} = conn, _params) do
+    username = auth.uid
+    expiration = auth.credentials.expires_at
+
+    current_time = DateTime.utc_now() |> DateTime.to_unix()
```
paulswartz: `current_time = System.system_time(:second)`

lib/signs_ui_web/controllers/auth_controller.ex
```diff
@@ -0,0 +1,25 @@
...
+
+    conn
+    |> Guardian.Plug.sign_in(
```
paulswartz: Does `uberauth_cognito` take care of verifying that the callback from Cognito is valid, or is that something we'd need to do here?

test/signs_ui_web/channels/user_socket_test.exs
```diff
@@ -0,0 +1,29 @@
...
+  use SignsUiWeb.ChannelCase
+
+  setup do
+    %{socket: socket()}
+  end
+
+  describe "connect/2" do
+    test "authenticates when a valid token is given", %{socket: socket} do
+      current_time = DateTime.utc_now() |> DateTime.to_unix()
+      expiration_time = current_time + 100
```
paulswartz: You might want to add more time to account for CI being slower.

test/signs_ui_web/channels/signs_channel_test.exs
```diff
@@ -0,0 +1,89 @@
+defmodule SignsUiWeb.SignsChannelTest do
+  use SignsUiWeb.ChannelCase
+  import ExUnit.CaptureLog
+
+  setup do
+    socket = subscribe_and_join!(socket(), SignsUiWeb.SignsChannel, "signs:all", %{})
+    %{socket: socket}
+  end
+
+  describe "handle_in" do
```
paulswartz: Could some of these use `@tag :authenticated`?

paulswartz:
Feedback on `ueberauth_cognito`: https://docs.google.com/document/d/11J1EPK7VaMeGfBdykKwsrrpAczveBQ3Ev7fJ28dERIA/edit?usp=sharing

test/signs_ui_web/channels/signs_channel_test.exs
lemald: The `@tag :authenticated` thing is implemented for conn case but not for channel case. We had a lot of trouble with implementing a setup block for channel cases that allows us to do a `@tag :authenticaed`-type thing, so these tests just use a combination of `encode_and_sign` and `assign_rts` if they need an authenticated socket.

lib/signs_ui_web/channels/signs_channel.ex
lemald: Fixed

lib/signs_ui_web/controllers/auth_controller.ex
lemald: Updated here and elsewhere in the PR where `DateTime.utc_now()` was used.

test/signs_ui_web/channels/user_socket_test.exs
lemald: Bumped this up to 500 seconds in this and the other tests where we have a future expiration time.

lib/signs_ui_web/controllers/auth_controller.ex
paulswartz: Looks like the answer is `uberauth_cognito` should verify the callback, but currently does not.

losvedir: Yeah, I agree it's `ueberauth_cognito`'s responsibility. What do you mean by verify the callback? There's a few parts to the process. I would say that it's mostly verified. We have a TODO on one part, which we do plan on doing, but I don't believe applies to our use case anyway.

When a request comes into signs-ui's `/auth/cognito/callback` route, it expects a `code=[UUID]` parameter. We then take that code, on the backend, and POST it to Cognito's `/oauth2/token` endpoint. The code must be valid, or else Cognito won't return the access token and JWT profile "bearer token".

The JWT we get back in that request includes a header/signature and could be validated, which we have a TODO on. However, at this point, I believe we _know_ it's valid, since we made the request ourselves to AWS over a TLS connection. As I understand it, the reason for the header/signature bit, is JWT's are meant to be able to be passed around from service to service, and any given service should be able to validate that it's authentic. In our scenario, we got it ourselves over TLS, so we know it's authentic, and immediately use Guardian to create/sign a new token which we put in the session.

One further validation/security mechanism that we have not currently implemented is PKCE. But this one, as I understand it, has a threat model relevant to native apps, where a malicious app running on the device can obtain the initial callback `code` and beat your own app to sending it to the oauth server to get the access token. While it's nice to have, I don't believe it applies to us, with a single app running in our virtualized container.

losvedir: Oh, if you mean the response we get back from Cognito's `/oauth2/token` endpoint, I just realized the version in the repo doesn't have a case clause for anything other than a successful response, so rather than the app getting an authentication failure it just blows up. I have that fixed in my local branch but hadn't pushed that up yet.

lemald:
> Feedback on `ueberauth_cognito`: https://docs.google.com/document/d/11J1EPK7VaMeGfBdykKwsrrpAczveBQ3Ev7fJ28dERIA/edit?usp=sharing

Hey Paul,

Thanks for the feedback. We believe we've now addressed all of the major feedback on `ueberauth_cognito` besides the stuff around validating the JWT from Cognito, which we're going to tackle after lunch. 🍕

lib/signs_ui_web/controllers/auth_controller.ex
paulswartz: Can you open a PR for that when it's going in? It's harder to review the code once it's in the master branch.

lemald: Good point, we'll do an actual PR for that.

config/config.exs
losvedir: Good idea. I created and configured an Ueberauth strategy for use in the dev and test environments that just logs you in, rather than redirecting away to AWS.

losvedir:
Okay, I think all the comments have been addressed here, and in the `ueberauth_cognito` app. This PR should be ready for another review.

lib/signs_ui_web/channels/signs_channel.ex
```diff
@@ -9,15 +9,53 @@ defmodule SignsUiWeb.SignsChannel do
...
   def handle_in("changeSigns", changes, socket) do
-    new_signs =
-      changes
-      |> Enum.map(fn {id, config} -> {id, Sign.from_config(id, config)} end)
-      |> Enum.into(%{})
+    if socket_authenticated?(socket) do
+      new_signs =
+        changes
+        |> Enum.map(fn {id, config} -> {id, Sign.from_config(id, config)} end)
+        |> Enum.into(%{})
```
paulswartz: `new_signs = Map.new(changes, fn {id, config} -> ... end)`

lib/signs_ui_web/channels/signs_channel.ex
```diff
@@ -9,15 +9,53 @@ defmodule SignsUiWeb.SignsChannel do
...
+      {:error, _error} -> false
+    end
+  end
+
+  @spec maybe_send_auth_expired_message(Phoenix.Socket.t()) :: Phoenix.Socket.t()
+  defp maybe_send_auth_expired_message(socket) do
+    if socket.assigns[:auth_expired_message_sent?] do
+      socket
+    else
+      :ok = push(socket, "auth_expired", %{})
```
paulswartz: Should the socket get disconnected at this point as well?

test/signs_ui/signs/state_test.exs
```diff
@@ -1,11 +1,23 @@
 defmodule SignsUI.Signs.StateTest do
-  use ExUnit.Case
-  use Phoenix.ChannelTest
+  use SignsUiWeb.ChannelCase
   import SignsUI.Signs.State
   alias SignsUI.Signs
   alias SignsUI.Signs.Sign
 
-  @endpoint SignsUiWeb.Endpoint
+  setup _tags do
```
paulswartz: If you aren't using the tags, you can do `setup do`

lib/signs_ui/ueberauth/strategy/fake.ex
```diff
@@ -0,0 +1,38 @@
+defmodule SignsUi.Ueberauth.Strategy.Fake do
+  use Ueberauth.Strategy
+
+  def handle_request!(conn) do
```
paulswartz: I think if you mark these as `@impl Ueberauth.Strategy` then you'll get compile time warnings if you don't implement all the functions.

test/signs_ui_web/controllers/messages_controller_test.exs
```diff
@@ -9,21 +9,30 @@ defmodule SignsUiWeb.MessagesControllerTest do
...
       assert html_response(conn, 200) =~ "MBTA Realtime Signs"
     end
+
+    @tag :authenticated
+    test "includes user token", %{conn: conn} do
+      conn = get(conn, messages_path(conn, :index))
+
+      response = html_response(conn, 200)
+
+      token = Regex.run(~r/window.userToken = "(.+)";/, response) |> Enum.at(1)
```
paulswartz: Should the token be available to Javascript? I thought the best practice here was to have the token as an HTTPOnly cookie, so that the browser is responsible for sending it but that it's not available to attackers?

lib/signs_ui_web/channels/signs_channel.ex
lemald: We experimented with having the `handle_in` / `handle_out` functions return `{:stop, ...}` in this case, but unfortunately that gets logged as a warning with no apparent way to reduce the severity. I suppose we could also close the socket on the front-end when it receives the `auth_expired` message, but it also calls `window.location.reload` in that handler so I'm not sure to what degree that would be an improvement.

lib/signs_ui_web/channels/signs_channel.ex
lemald: Done

test/signs_ui/signs/state_test.exs
lemald: Done

lib/signs_ui/ueberauth/strategy/fake.ex
lemald: Done

test/signs_ui_web/controllers/messages_controller_test.exs
losvedir: This is the approach [recommended by Chris McCord](https://elixirforum.com/t/accessing-cookies-in-phoenix-socket-connect/4731/4) in a question about using HTTPOnly cookies with websockets. Phoenix by design doesn't give access to the cookie in the socket/channel callbacks and recommends passing a signed `Phoenix.Token` via JS like we're doing. He says that it's more secure this way anyway, since websockets aren't subject to the same-origin cookie policy that http requests are.

lib/signs_ui_web/channels/signs_channel.ex
losvedir: Yeah, as @lemald mentions, returning `{:stop, ...}` here was our first approach, but that terminated the channel and socket GenServers noisily in the logs. Perhaps we could go down the route of trapping exits and handling the termination messages? We didn't explore it too much since at this point they're not able to make any changes over the websocket and we wanted to send the frontend a message anyway, to trigger them to reload the page (and get redirected to Cognito).

paulswartz: What was the full return you used? I'd figure `{:stop, :normal, _state}` would close the socket normally.

losvedir: Oh, you know what, I think we were using `{:stop, :authentication_expired}` or something to that effect. The Phoenix docs for the `handle_in` and `handle_out` callbacks don't mention anything about stopping normally, but you're right that the normal GenServer conventions probably apply. We'll try that.

test/signs_ui_web/controllers/messages_controller_test.exs
paulswartz: Thanks: I didn't know about that attack, so I learned something today!

mix.exs
```diff
@@ -49,11 +49,15 @@ defmodule SignsUi.Mixfile do
...
+      {:plug_cowboy, "~> 1.0"},
       {:poison, "~> 3.1"},
       {:httpoison, "~> 1.0"},
       {:ex_aws, "~> 2.0"},
       {:ex_aws_s3, "~> 2.0"},
-      {:timex, "~> 3.1"}
+      {:timex, "~> 3.1"},
+      {:guardian, "~> 1.0"},
+      {:ueberauth, "~> 0.1"},
+      {:ueberauth_cognito, git: "https://github.com/mbta/ueberauth_cognito.git"}
```
paulswartz: You should release this at some point. Another notch in our open-source belt!

lib/signs_ui_web/channels/signs_channel.ex
losvedir: After a quick test, you're right: `{:stop, :normal, socket}` closes it normally without errors in the logs. We changed to that.

mix.exs
losvedir: Yep, we plan to! We have a ticket in our backlog for that.

PR #90, opened 2019-06-19
Use groups in authentication

lemald:
#### Summary of changes
**Asana Ticket:** [🕵️ Levels of permission](https://app.asana.com/0/584764604969369/1124399243620929/f)

Takes the group information from Ueberauth and includes it in the Guardian token as well. When trying to access the viewer, go through a new plug that verifies that the user has the appropriate group. If not, return a 403 with an accompanying message. In the future we can add a `signs-ui-readonly` group or something like that by generalizing this somewhat (will also require tweaks to how the token for the websocket works, I think).

Instead of writing a new plug, I could have used `plug(Guradian.Plug.EnsureAuthenticated, %{groups: ["signs-ui-admin"]})` in tandem with updated to the error handler to somehow be aware of what kind of check we're doing against the token and then either redirecting to Cognito if it's completely missing, or giving the 403 is the group isn't present. That seemed a bit too underhanded to me, though, so I decided to be explicit and create a separate plug.

#### Reviewer Checklist

### DIFF
--- a/lib/signs_ui_web/channels/signs_channel.ex
+++ b/lib/signs_ui_web/channels/signs_channel.ex
@@ -40,9 +40,11 @@ defmodule SignsUiWeb.SignsChannel do
     claims = Guardian.Phoenix.Socket.current_claims(socket)
     token = Guardian.Phoenix.Socket.current_token(socket)
 
-    case SignsUiWeb.AuthManager.decode_and_verify(token, claims) do
-      {:ok, _claims} -> true
-      {:error, _error} -> false
+    with {:ok, claims} <- SignsUiWeb.AuthManager.decode_and_verify(token, claims),
+         true <- SignsUiWeb.AuthManager.claims_grant_signs_access?(claims) do
+      true
+    else
+      _ -> false
     end
   end
 
--- a/lib/signs_ui_web/ensure_signs_ui_group.ex
+++ b/lib/signs_ui_web/ensure_signs_ui_group.ex
@@ -6,7 +6,7 @@ defmodule SignsUiWeb.EnsureSignsUiGroup do
   def call(conn, _opts) do
     claims = Guardian.Plug.current_claims(conn)
 
-    if not is_nil(claims["groups"]) and @signs_ui_group in claims["groups"] do
+    if SignsUiWeb.AuthManager.claims_grant_signs_access?(claims) do
       conn
     else
       conn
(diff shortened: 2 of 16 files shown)


lib/signs_ui_web/ensure_signs_ui_group.ex
```diff
@@ -0,0 +1,25 @@
+defmodule SignsUiWeb.EnsureSignsUiGroup do
+  import Plug.Conn
+
+  @signs_ui_group "signs-ui-admin"
+
+  @unauthorized_message "You are not authorized to access the signs viewer. " <>
```
phildarnowsky: @lemald I'm not wild about having this big blob of HTML in a module that does authorization logic, is there perhaps an idiomatic way in Phoenix to render a template from within a random plug like this?

paulswartz: You can raise an exception, and then have ErrorView render a response for it.

losvedir:
I think there should be some authorization at the websocket level, too.

lemald:
@losvedir hmm, good point, I forgot we use the same token for the websocket now. I'll add that.

lib/signs_ui_web/ensure_signs_ui_group.ex
lemald: I'm not sure if `ErrorView` is quit the correct place to put it (I don't want it to trigger the stacktrace thing-y in development mode), but I did create an actual view and such so it's no longer just a constant in this module.

lib/signs_ui_web/channels/signs_channel.ex
```diff
@@ -44,6 +44,13 @@ defmodule SignsUiWeb.SignsChannel do
       {:ok, _claims} -> true
       {:error, _error} -> false
     end
+
+    with {:ok, claims} <- SignsUiWeb.AuthManager.decode_and_verify(token, claims),
```
losvedir: Maybe make the `case` expression above part of this `with` chain?

losvedir: Er, actually I think shouldn't it just be deleted?

lemald: Oops!

lib/signs_ui_web/templates/unauthorized/index.html.eex
```diff
@@ -0,0 +1,3 @@
+You are not authorized to access the signs viewer.
+If you believe you should have access, please have your supervisor contact
+<a href=\"realtimeappsdl@mbta.com\">realtimeappsdl@mbta.com</a>.
```
losvedir: The backslashes are superfluous now. Also, I don't think this should be a link anyway. Maybe a `mailto:` but actually probably not since this person isn't the one doing the mailing.

lemald: Fixed.

PR #96, opened 2019-07-12
Handle refresh tokens

lemald:
#### Summary of changes
**Asana Ticket:** [👁️ handle token refreshes](https://app.asana.com/0/584764604969369/1123278898060154/f)

1. Created a new GenServer that stores user refresh tokens on the backend. Currently it's just a map in-memory but could use some sort of persistent storage in the future.
2. Store the username in the session, since the Guardian functions don't return any information once the token is expired and we need the username to lookup the refresh token.
3. Upon using the refresh token, clear it from the GenServer state. This is to prevent a loop if it turns out that the token is expired.
4. When we get a failure from Ueberauth, only return a 403 if we weren't trying to use the refresh token. If it was a failure with the refresh token, instead redirect to the normal start of the workflow to allow the user to log back in.

#### Reviewer Checklist

### DIFF
--- a/lib/signs_ui/refresh_token_store.ex
+++ b/lib/signs_ui/refresh_token_store.ex
@@ -1,0 +1,34 @@
+defmodule SignsUi.RefreshTokenStore do
+  @moduledoc """
+  Stores refresh tokens for user authentication. Currently these are
+  just kept in memory because there's only one instance of the
+  application running and deploys / restarts are rare, but if need be
+  the data could be moved into persistent storage in the future.
+  """
+
+  use Agent
+
+  @type state :: %{
+          String.t() => String.t()
+        }
+
+  def start_link(opts \\ []) do
+    name = opts[:name] || __MODULE__
+    Agent.start_link(fn -> %{} end, name: name)
+  end
+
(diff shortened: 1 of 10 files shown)


lib/signs_ui/refresh_token_store.ex
```diff
@@ -0,0 +1,50 @@
...
+
+  use GenServer
+
+  @type state :: %{
+          String.t() => String.t()
+        }
+
+  def start_link(opts \\ []) do
+    name = opts[:gen_server_name] || __MODULE__
+    GenServer.start_link(__MODULE__, %{}, name: name)
```
losvedir: Since this is basically a simple key-value store, I wonder if it's any clearer if this module is implemented as an `Agent` (or ETS, I guess)?

lemald: Huh, I briefly considered ETS but I haven't actually used Agent before, so it didn't occur to me. I'll look into it (or maybe just use ETS, honestly, since I'm already familiar with it).

lemald: I ended up using Agent, it made the code somewhat simpler.

PR #370, opened 2020-11-02
feat: chelsea bridge announcements toggle

fjlanasa:
#### Summary of changes
**Asana Ticket:** [👁 transitway engine / signs-ui: Chelsea bridge toggle](https://app.asana.com/0/584764604969369/1198221727983513/f)

Changes:
* Added `chelsea_bridge_announcements: "auto" | "off"` to config
* Added logic to parse and return this config on initial load
* Added `changeChelseaBridgeAnnouncements` handler for updates
* Added toggle component to `Line.jsx` to handle updating 

I deployed this to dev, saw that toggling the switch on/off changed the config appropriately, then redeployed master, which is working as expected. 

#### Reviewer Checklist


assets/js/Line.jsx
```diff
@@ -95,6 +97,26 @@ function Line({
           readOnly={readOnly}
         />
       )}
+      {line === 'Silver' && (
+        <div>
+          <label htmlFor="chelsea_boarding">
```
digitalcora, line 102: Since the input is nested inside the label, I don't think we need a `for`/`id` linkage here.

assets/js/Line.test.js
```diff
@@ -414,3 +414,81 @@ test('Sign config is not affected by batch updates if sign does not support mode
...
+
+  const currentTime = now + 2000;
+  const line = 'Silver';
+  const signConfigs = {};
+  const setConfigs = jest.fn(() => true);
+  const chelseaBridgeAnnouncements = 'off';
+  const setChelseaBridgeAnnouncements = jest.fn(() => true);
+  const setConfiguredHeadways = () => {};
+  const readOnly = false;
+  const configuredHeadways = {};
```
digitalcora, line 430: Aside from `setChelseaBridgeAnnouncements`, it doesn't seem like we use any of these variables beyond the `React.createElement` that comes immediately after. Could we consolidate them? I think this might be more readable since it would take up much less vertical space.

```js
React.createElement(
  Line,
  {
    signs: {},
    currentTime: Date.now() + 2000,
    line: 'Silver',
    // ...
  }
)
```

lib/signs_ui_web/channels/signs_channel.ex
```diff
@@ -50,6 +50,27 @@ defmodule SignsUiWeb.SignsChannel do
     end
   end
 
+  def handle_in("changeChelseaBridgeAnnouncements", changes, socket) do
+    case socket_access_level(socket) do
```
digitalcora, line 54: Some of these access control bits now appear duplicated 3 times in this module — maybe worth an extraction? Like:

```elixir
with_admin_access(socket, fn ->
  {:ok, _new_state} = SignsUi.Config.State.update_chelsea_bridge_announcements(changes)
  username = Guardian.Phoenix.Socket.current_resource(socket)

  Logger.info(
    "chelsea_bridge_announcements_changed: user=#{username} changes=#{inspect(changes)}"
  )
end)
```

This could have a better name to more clearly imply what it does, but this is the best one I could think up at the moment.

assets/js/Line.jsx
```diff
@@ -95,6 +97,26 @@ function Line({
...
+            Boarding Announcements
```
digitalcora, line 103: Shouldn't this be "Bridge" Announcements? (it looks like there are also some parts of the code that inconsistently use "boarding" vs. "bridge")

lib/signs_ui/config/state.ex
```diff
@@ -47,6 +56,11 @@ defmodule SignsUi.Config.State do
         save_sign_config_changes(%{}, config)
         save_configured_headways_changes(config.configured_headways, config)
 
+        save_chelsea_bridge_announcements(
+          Map.get(config, :chelsea_boarding_announcements, "off"),
```
digitalcora, line 60: Is this supposed to be "boarding"? I can't seem to find the corresponding test, if there is one.

fjlanasa: yeah, should be

assets/js/Line.jsx
fjlanasa: yeah, will change

lib/signs_ui/config/state.ex
digitalcora: Hm, why didn't any tests fail when this value was incorrect? I guess we must have test coverage of this line in some way or else Codecov would have complained, but maybe we're not checking the actual value of the saved config.

assets/js/Line.jsx
fjlanasa: yeah, i'd like to leave the `id` (mainly for testing), and it seems like no harm to leave the `htmlFor` here in that case.

digitalcora: For the testing, rather than `find("input#chelsea_boarding")` does something like `find("input[name=chelsea_boarding]")` work? I think it's fine if we leave the ID, of course, this is more curiosity 🙂

fjlanasa: i'm sure it does.

lib/signs_ui_web/channels/signs_channel.ex
fjlanasa: 👍 pushed up this change. looks good.

lib/signs_ui_web/channels/signs_channel.ex
```diff
@@ -77,4 +71,20 @@ defmodule SignsUiWeb.SignsChannel do
     :ok = push(socket, "auth_expired", %{})
     socket
   end
+
+  @spec with_admin_access(Phoenix.Socket.t(), fun()) :: {:noreply, Phoenix.Socket.t()}
```
digitalcora, line 75: I think the return type should be:

```elixir
{:noreply, Phoenix.Socket.t()} | {:stop, :normal, Phoenix.Socket.t()}
```

Unfortunately it seems like there is no handy type alias in `Phoenix.Channel` for the type of things that should be returned from `handle_in`.

lib/signs_ui/config/state.ex
fjlanasa: the logic was tested [here](https://github.com/mbta/signs_ui/blob/fl/chelsea/test/signs_ui/config/state_test.exs#L162-L163), and happened to pass because of the fallback to `"off"` in the `Map.get` statement above. I flipped the order of the assertions, which caused the test to fail before I changed the key.

digitalcora (approved):
🍰

PR #472, opened 2021-03-24
feat: add logging with Sentry

bklebe:
#### Summary of changes
**Asana Ticket:** [💳 Add sentry to signs-UI](https://app.asana.com/0/584764604969369/1198979612069229/f)

Added Sentry v8 to signs_ui, including the logger backend and the plug in `SignsUiWeb.Endpoint`, as advised by the Sentry documentation. I also had to update `plug_cowboy` and `cowboy` to version 2, as they were incompatible.

#### Reviewer Checklist


lib/utilities/config.ex
```diff
@@ -4,4 +4,10 @@ defmodule SignsUi.Utilities.Config do
       Application.put_env(:signs_ui, key, val)
     end
   end
+
+  def update_env(app, key, val) do
```
digitalcora, line 8: We should be able to combine these into a single definition:

```elixir
def update_env(app \\ :signs_ui, key, val) do
```

config/config.exs
```diff
@@ -28,13 +28,21 @@ config :logger, :console,
...
 config :ueberauth, Ueberauth,
   providers: [
     cognito: {SignsUi.Ueberauth.Strategy.Fake, []}
   ]
 
 config :phoenix, :json_library, Jason
 
+config :sentry,
+  enable_source_code_context: true,
+  root_source_code_path: File.cwd!(),
```
digitalcora, line 43: Should this not be `root_source_code_paths: [File.cwd!()]`? I don't see a "singular" version of this option in the documentation.

digitalcora (commented):
General comments:

1. In the [Phoenix setup instructions](https://github.com/getsentry/sentry-elixir#setup-with-plug-and-phoenix) it seems like we're also supposed to add a `use Sentry.PlugCapture` before the `use Phoenix.Endpoint`.

2. In the config, we can add an `in_app_module_allow_list` to tell Sentry which modules are part of our app, to improve the legibility of stack traces. I think here the list could just be `[SignsUi, SignsUiWeb]`.

3. Can we do a test deploy of this branch to `dev` before merging it, and check that a crash or error shows up in Sentry? If there's no way to intentionally generate one currently, we could temporarily add a commit that e.g. makes visiting a certain path always crash.

config/config.exs
bklebe: I got that code from the docs it presented me when I created the project.

digitalcora: Ah, it appears the singular version still works, [but is deprecated](https://github.com/getsentry/sentry-elixir/blob/7a1ba52c1ffd1c3b0371f9a00ed172e4cfabca74/lib/sentry/config.ex#L91).

bklebe:
@digitalcora 
> 1. In the [Phoenix setup instructions](https://github.com/getsentry/sentry-elixir#setup-with-plug-and-phoenix) it seems like we're also supposed to add a `use Sentry.PlugCapture` before the `use Phoenix.Endpoint`.

Right, my mistake.

> 2. In the config, we can add an `in_app_module_allow_list` to tell Sentry which modules are part of our app, to improve the legibility of stack traces. I think here the list could just be `[SignsUi, SignsUiWeb]`.

Done 👍 

> 3. Can we do a test deploy of this branch to `dev` before merging it, and check that a crash or error shows up in Sentry? If there's no way to intentionally generate one currently, we could temporarily add a commit that e.g. makes visiting a certain path always crash.

Currently running in dev. They suggest a testing method here, which I'll try locally: https://github.com/getsentry/sentry-elixir#testing-your-configuration

config/config.exs
```diff
@@ -28,13 +28,22 @@ config :logger, :console,
...
 config :ueberauth, Ueberauth,
   providers: [
     cognito: {SignsUi.Ueberauth.Strategy.Fake, []}
   ]
 
 config :phoenix, :json_library, Jason
 
+config :sentry,
+  enable_source_code_context: true,
+  root_source_code_path: [File.cwd!()],
```
digitalcora, line 43: The key should be plural: `root_source_code_paths`. (this is still the deprecated singular form)

bklebe: 🤦‍♀️ my b

mix.exs
```diff
@@ -48,8 +48,8 @@ defmodule SignsUi.Mixfile do
       {:phoenix_html, "~> 2.10"},
       {:phoenix_live_reload, "~> 1.0", only: :dev},
       {:gettext, "~> 0.11"},
-      {:cowboy, "~> 1.0"},
-      {:plug_cowboy, "~> 1.0"},
+      {:cowboy, "~> 2.7"},
```
digitalcora, line 51: I remembered something about this, so I went and checked, and it seems like explicitly depending on `cowboy` is not needed in general for Phoenix apps; the [1.4 upgrade guide](https://gist.github.com/chrismccord/bb1f8b136f5a9e4abc0bfc07b832257e) recommends removing it ("_replace_ your cowboy dependency with plug_cowboy").

digitalcora (approved):
I had a small comment about dependencies. In any case, if we have added the env vars to both dev and prod, seems good to merge!

PR #492, opened 2021-04-14
Alert expiration preparation

losvedir:
#### Summary of changes
**Asana Ticket:** [🔀 Signs-UI is prepared for alerts](https://app.asana.com/0/281253957434160/1200158594431786)

If you give a mouse a cookie... I wanted to use the nice new standard library functions `DateTime.new` and `Calendar.strftime`, but those require Elixir 1.11, so there's a commit to upgrade Elixir and Erlang (and switch semaphore to using asdf like our other repos). But then Elixir 1.11 makes something in a Phoenix macro a warning (about syntax being inconsistent between a function call and a map), which because of `--warnings-as-errors` makes CI fail (since the macro makes it look like it happens in our own Router code), so then I had to update Phoenix to 1.5, too. Finally, since we're then on the newest Elixir with all the best standard library datetime functions, I ripped out Timex.

And then after that, there's a commit which is the actual point of this PR, which lays the foundation for the alerts expiration work. It creates a new GenServer in our Application supervision tree, `SignsUi.Alerts.State`, which adds or removes an alert every 15 seconds, and then broadcasts it to the frontend. While there's tests for some things that are real (like JSON serialization of the new Alert struct), I also have only more minimal tests for the actual `SignsUi.Alerts.State` to hit our code coverage. I tested the API streaming and it worked well, so I figure that's what the real implementation will use; but that also means I'm not so familiar with how the implementation and testing will look, so I didn't want to scaffold up something that might be inappropriate.

#### Reviewer Checklist
- [x] Meets ticket's acceptance criteria
- [x] Any new or changed functions have typespecs
- [x] Tests were added for any new functionality (don't just rely on Codecov)


lib/signs_ui/config/sign.ex
```diff
@@ -97,49 +102,8 @@ defmodule SignsUi.Config.Sign do
         mode: :static_text,
         line1: config["line1"],
         line2: config["line2"],
-        expires: expiration_from_string(config["expires"])
-      }
-    }
-  end
-
-  @spec from_config(String.t(), map()) :: t()
```
losvedir, line 105: The implementation here is identical to `from_json` above it, and has no tests and was only used in one place. I don't really remember how or why it's here... and sort of suspect it might have been a merge conflict or something back in the day? In any case, I removed it, and switched its one caller to `from_json`.

mkennedm (approved):
Looks good.

lib/signs_ui/messages/sign_content.ex
```diff
@@ -52,15 +52,15 @@ defmodule SignsUi.Messages.SignContent do
   defparsec(:parse_command, command)
 
   @spec new(String.t(), String.t(), DateTime.t()) :: {:ok, t()} | {:error, any()}
-  def new(station, command, now \\ Timex.now()) when is_binary(station) do
+  def new(station, command, now \\ DateTime.now!("Etc/UTC")) when is_binary(station) do
```
digitalcora, line 55: `DateTime.utc_now()` I think would be the preferred way to do this (here and in the various test files)

lib/signs_ui/alerts/alert.ex
```diff
@@ -0,0 +1,44 @@
...
+          created_at: format_dt(value.created_at),
+          service_effect: value.service_effect
+        },
+        opts
+      )
+    end
+
+    @spec format_dt(DateTime.t() | nil) :: String.t()
+    defp format_dt(nil) do
+      ""
```
digitalcora, line 26: Can an alert's `created_at` ever be `nil`? I thought this was a non-null field.

lib/signs_ui/alerts/alert.ex
```diff
@@ -0,0 +1,44 @@
...
+      date =
+        if DateTime.to_date(dt) == today do
+          "Today"
+        else
+          Calendar.strftime(dt, "%m/%d")
+        end
+
+      time = Calendar.strftime(dt, "%-I:%M %p")
+
+      "#{date} @ #{time}"
```
digitalcora, line 41: This seems like we're embedding presentational logic in the backend data model. Can we emit this in the JSON as ISO8601, and have the React app format it? (as with e.g. `expiration_to_iso8601`)

lib/signs_ui/alerts/state.ex
```diff
@@ -0,0 +1,63 @@
+defmodule SignsUi.Alerts.State do
+  use GenServer
+  require Logger
+  alias SignsUi.Alerts.Alert
+
+  defstruct alerts: %{}
+
+  @type route_id :: String.t()
+  @type alert_id :: String.t()
```
digitalcora, line 9: This type appears unused (we do have `Alert.id()` so it seems redundant).

lib/signs_ui/alerts/state.ex
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+    SignsUiWeb.Endpoint.broadcast!("signs:all", "new_alert_state", new_alert_state)
+    Logger.info("new_alert_state: #{inspect(new_alert_state)}")
+
+    {:noreply, put_in(state.alerts, new_alert_state)}
+  end
+
+  def handle_info(msg, state) do
+    Logger.info("#{__MODULE__} unknown_message #{inspect(msg)}")
```
digitalcora, line 60: Maybe this should at least be a `warn` — we don't expect it to ever happen except for that Hackney bug, so if it happens in development/test you might want to have your attention called to it.

test/signs_ui/alerts/state_test.exs
```diff
@@ -0,0 +1,24 @@
+defmodule SignsUi.Alerts.StateTest do
+  use ExUnit.Case, async: true
+  use SignsUiWeb.ChannelCase
+
+  describe "start_link" do
+    test "GenServer runs without crashing" do
+      @endpoint.subscribe("signs:all")
+
+      {:ok, _pid} = SignsUi.Alerts.State.start_link(interval_ms: 50)
+      Process.sleep(70)
```
digitalcora, line 10: I don't think we should need this — `assert_broadcast` by default waits up to the ExUnit `assert_receive` timeout (which in turn defaults to 100ms) for the given message to be broadcasted. If we need to wait longer, we can also give a third argument to `assert_broadcast`.

test/signs_ui/alerts/state_test.exs
```diff
@@ -0,0 +1,24 @@
...
+
+      assert_broadcast("new_alert_state", %{})
+    end
+  end
+
+  describe "handle_info" do
+    test "handles unknown messages without crashing" do
+      {:ok, pid} = SignsUi.Alerts.State.start_link()
+      send(pid, :unknown!)
+      Process.sleep(50)
```
digitalcora, line 20: Rather than sleeping, since we have a reference to the process, we could use `Process.monitor` and `refute_receive` on the DOWN message.

lib/signs_ui/alerts/alert.ex
losvedir: I was trying to avoid having to pull in `moment` or some other big JS library for time zone conversions (since "today" has to be time zone aware). But good catch about how we already are sending over `iso8601` - I forgot the browser already handles locales for us. Yeah, I'll make this switch. In fact, it means I don't need to prove my own `defimpl Jason.Encoder` anymore for `Alert` and can just `@derive` it.

lib/signs_ui/alerts/alert.ex
losvedir: Well, the API documentation calls it non-null, and same for `service_effect`, but that one, too, I opted for `| nil`. But to me, from the point of view of signs-ui, it's outside its application boundary, so it shouldn't just rely on that assertion. If an alert somehow ends up without a `created_at`, that's not an excuse for signs_ui to go down.

So that implies to me that if we want to internally enforce a model of a required `created_at`, in the later PR when we actually are dealing with alerts, we'd be forced to drop any alerts that are missing that field. But that piece of information didn't seem so critical to any of the functionality that I thought it would warrant that action. If you asked, "hypothetically, if we received an alert without a created_at, would we still want PIOs to be able to select it?" I think the answer is yes, right?

You could make the argument that a missing `created_at` implies data corruption or something, and so we _shouldn't_ show it... but I don't know if that's the case. I actually don't know what the API would do, but I don't at all trust alerts-ui to be robust software that always provides a `created_at` even if it claims to. I suspect the API would still pass that alert along, but I haven't verified that.

losvedir: Hmm, actually no I'm wrong. If the API received an alert missing a `created_at` it [looks like](https://github.com/mbta/api/blob/7dc7446dbcf6e833982b4aba392b2e3aad203b87/apps/parse/lib/parse/alerts.ex#L56) it would 💥 . Given that it hasn't... I guess in this particular way alerts-ui has been pretty consistent at least. I still don't think it's worth signs_ui putting its foot down and demanding its presence here, though.

PR #496, opened 2021-04-22
Update frontend

losvedir:
#### Summary of changes
**Asana Ticket:** n/a

A bit of frontend tweaks before I dive in with the actual react alerts feature.

The first commit switches to npm, since that tends to be more common in CTD (based on a look through our repos), and in particular is what screens uses. I used yarn back when signs_ui started because it was dramatically better than npm at the time. But I think now npm has closed the gap somewhat, and the new web apps since then (skate, screens, arrow) use the phoenix default of npm. Rather than dealing with translating yarn.lock to package-lock.json, I took this opportunity to do one of our dependabot-style frontend dep refreshes. I updated `package.json` to have `^` constraints wherever they were pinned to a specific version, so that the fresh `npm install` would resolve to the latest patch versions of all our dependencies.

The second commit removes some CSS that no longer applies. We used to have a simple CRUD checkbox based sign control page before the current react-app. When I removed that (in #70) I missed this file.

The third commit resolves some browser differences in font size, which caused the sign preview to overflow. The font-size was set at a relative `em` but a base font size was never set so different browsers computed a different font-size in px. This 1) sets a base font size on the `body`, changes the font size of the signs to `rem` (which is what Frank did in all his frontend work, so... I'm assuming that's a good practice), and ensures the text fits on the sign by changing the width to use `ch` since semantically what we're trying to do is fit a fixed number of characters on there.

#### Reviewer Checklist


assets/__mocks__/popper.js.ts
```diff
@@ -0,0 +1,50 @@
...
+          in div (created by r)
+          in r (created by InnerPopper)
+          in InnerPopper (created by Context.Consumer)
+          in Popper (created by r)
+          in Manager (created by r)
+          in r (created by r)
+          in r (created by SetExpiration)
+          in div (created by SetExpiration)
+          in SetExpiration (created by WrapperComponent)
+          in WrapperComponent
```
digitalcora, line 28: We probably don't need to include this full "element trace", especially since it would be different for every test.

semaphore/setup.sh
```diff
@@ -24,4 +24,4 @@ mix local.hex --force
 mix local.rebar --force
 mix deps.get
 
-yarn --cwd assets install
+npm --prefix assets install
```
digitalcora, line 27: `npm install --prefix assets` might read better?

assets/css/viewer.css
```diff
@@ -43,10 +43,11 @@
 .viewer--sign-line {
   color: rgb(251, 220, 75);
   font-family: monospace;
-  font-size: 1.2em;
+  font-size: 1rem;
```
digitalcora, line 46: Is this needed to override another font size somewhere between the body and this element? Without a specified font size it would already be `1rem`, by definition.

digitalcora (approved):
Pending confirmation that you've done a successful test deploy of this branch (for the Docker changes), all looks good to me!

PR #508, opened 2021-04-26
feat: Sign mode is changed to auto when alert is inactive

mkennedm:
#### Summary of changes
**Asana Ticket:** [🔀 Signs-UI expires custom text when alert becomes inactive](https://app.asana.com/0/584764604969369/1200158594431784/)

`expire_signs_via_alert` takes a list of `sign_states` and a `MapSet` of `alert_ids` and returns a list of `SignsUi.Config.Sign`. Each `Sign` has an `id` that is not in `alert_ids` and a `config.mode` set to `:auto`.

`alert_ids` is a `MapSet` of active (not expired) `alert_id`s.

#### Reviewer Checklist


lib/signs_ui/config/expiration.ex
```diff
@@ -59,6 +59,59 @@ defmodule SignsUi.Config.Expiration do
...
+  end
+
+  @spec get_expired_sign_states(
+          list(SignsUi.Config.State.t()),
+          MapSet.t(),
+          list(SignsUi.Config.State.t())
+        ) :: list(SignsUi.Config.State.t())
+  def get_expired_sign_states(remaining_sign_states, alert_ids, expired_sign_states) do
+    if remaining_sign_states == [] do
+      expired_sign_states
```
paulswartz, line 82: can this be part of a pattern match in the function head?
```elixir
def expired_sign_states(sign_states, alert_ids, expired_signs)
def expired_sign_states([], _, expired_signs) do
  expired_signs
end
def expired_sign_states([sign_state | sign_states], alert_ids, expired_signs) do
  ...
end
```

lib/signs_ui/config/expiration.ex
```diff
@@ -59,6 +59,59 @@ defmodule SignsUi.Config.Expiration do
...
+          get_expired_sign_states(tail, alert_ids, expired_sign_states)
+
+        Enum.member?(alert_ids, sign_state.config.alert_id) ->
+          get_expired_sign_states(tail, alert_ids, expired_sign_states)
+
+        true ->
+          new_sign_state = %SignsUi.Config.Sign{
+            id: sign_state.config.alert_id,
+            config: %{mode: :auto}
+          }
```
paulswartz, line 97: Is there a reason to create a new `Sign` vs updating the existing one?
```elixir
sign_state = %{sign_state | config:  %{mode: :auto}}
```

lib/signs_ui/config/expiration.ex
```diff
@@ -59,6 +59,59 @@ defmodule SignsUi.Config.Expiration do
...
+
+    expired_sign_states
+  end
+
+  @spec get_expired_sign_states(
+          list(SignsUi.Config.State.t()),
+          MapSet.t(),
+          list(SignsUi.Config.State.t())
+        ) :: list(SignsUi.Config.State.t())
+  def get_expired_sign_states(remaining_sign_states, alert_ids, expired_sign_states) do
```
paulswartz, line 80: This can probably be a private function, since it's an implementation detail of `expire_signs_via_alert`.

lib/signs_ui/config/expiration.ex
```diff
@@ -59,6 +59,59 @@ defmodule SignsUi.Config.Expiration do
     |> Enum.into(%{})
   end
 
+  @spec expire_signs_via_alert(list(SignsUi.Config.State.t()), MapSet.t()) ::
```
paulswartz, line 62: Should this have the same shape as `expire_signs_via_time`?

lib/signs_ui/config/expiration.ex
```diff
@@ -59,6 +59,59 @@ defmodule SignsUi.Config.Expiration do
...
+
+          expired_sign_states = [new_sign_state | expired_sign_states]
+          get_expired_sign_states(tail, alert_ids, expired_sign_states)
+      end
+    end
+  end
+
+  @spec get_active_alert_ids(SignsUi.Alerts.State.state()) ::
+          MapSet.t(String.t())
+  def get_active_alert_ids(alert_state) do
```
paulswartz, line 107: In general, you don't need the `get_` prefix: the name of the function indicates what you're getting.

paulswartz (changes requested):
Can you add some higher-level tests? I see the unit tests, but not how the relevant alert data gets into the functions, or how `expire_signs_via_alert` is called.

lib/signs_ui/config/expiration.ex
paulswartz: There's also an existing `expire_single_sign` function you could use.

lib/signs_ui/config/expiration.ex
mkennedm: I don't think so. Here's what @losvedir said he wanted in Slack messages to me.
"a pure function called `expire_signs_via_alerts(sign_state, alert_ids)` where the alert_ids are the MapSet above, and the sign_state looks like: `[%SignsUi.Config.Sign{...}, ...]` ...  `expire_signs_via_alerts([signs...], alert_ids)` should return a list of the signs that have an alert_id which is not present in alert_ids. and those signs should be new signs with the same ID, but in auto mode"

lib/signs_ui/config/expiration.ex
mkennedm: I'm creating a new sign because that's what @losvedir asked for. "`expire_signs_via_alerts([signs...], alert_ids)` should return a list of the signs that have an `alert_id` which is not present in `alert_ids`. and those signs should be new signs with the same ID, but in auto mode"

test/signs_ui/config/expiration_test.exs
```diff
@@ -50,12 +50,139 @@ defmodule SignsUi.Config.ExpirationTest do
...
+            alert_id: "593"
+          }
+        }
+      ]
+
+      expired_signs = SignsUi.Config.Expiration.expire_signs_via_alert(sign_state, alert_ids)
+
+      expected_signs = [
+        %SignsUi.Config.Sign{
+          id: nil,
```
losvedir, line 165: These are supposed to be the sign IDs, not the alert IDs, of the signs that are getting expired (i.e.: signs that have an alert ID that is not present in the list of active alerts).

losvedir:
Okay, great start. I slacked you a couple pure functions to work on, to get the hang of the shape of the data and what exactly we're trying to do here. Now it's time to build this into the final PR.

## Update to Alerts GenServer

The "expiration" function uses the currently active alerts to know which signs should expire and revert back to "auto" mode. How do we know what the currently active alerts are? The `SignsUi.Alerts.State` GenServer. Right now that's being worked on in another ticket, to actually use the API and fill with real data, but the GenServer is in the code base and you can update it now to how you need it. You'll need to provide a client function that returns the alert_ids:

```ex
# to be placed below init/1

@spec active_alert_ids(GenServer.server()) :: [Alert.id()]
def active_alert_ids(pid \\ __MODULE__) do
  GenServer.call(pid, :active_alert_ids)
end
```

That does a "call" on the GenServer, passing `:active_alert_ids`, so you'll need to write the callback handler:

```ex
# to be placed above the first handle_info/2

def handle_call(:active_alert_ids, _from, state) do
  alert_ids = # ... code similar to your get_active_alerts function

  # the :reply means the second element of the tuple will be returned by the GenServer.call,
  # and the third element of the tuple is the GenServer's new state. In this case, it should be
  # the same as the old state.
  {:reply, alert_ids, state}
end
```

You can then test that in `tests/signs_ui/alerts/state_test.exs`. Add a new `describe` above the one for `handle_info`, with a test for your new function:

```ex
describe "handle_call :active_alert_ids" do
  test "returns the alert_ids out of the state" do
    state = %SignsUi.Alerts.State{
      alerts: %{
        # fill in sample data here, but it's of this shape:
        "Red" => %{
          "alert_id1" => %{},
          "alert_id2" => %{}
        },
        "Blue" => %{
          "alert_id3" => %{}
        }
      }
    }

    assert SignsUi.Alerts.State.handle_call(:active_alert_ids, self(), state) ==
      {:reply, MapSet.new(["alert_id1", "alert_id2", "alert_id3"]), state}
  end

  test "safely returns an empty map set if there are no alerts" do
    state = %SignsUi.Alerts.State{ alerts: %{} }
    assert SignsUi.Alerts.State.handle_call(:active_alert_ids, self(), state) ==
      {:reply, MapSet.new([]), state}
  end
end
```

## SignsUi.Config.Expiration gets the alert data

Now that `SignsUi.Alerts.State` is up and running, we need to update `SignsUi.Config.Expiration` to use that data, but also be "mockable" so that we can put whatever data we want in there for the purposes of our tests.

We can follow the convention that already exists in this GenServer for the `time_fetcher`, which fetches the current time in production, but also lets us inject in whatever time we want for tests.

There are a couple things to do:

Update the `@type state`, to include a new `:alert_fetcher` key, which is a function of zero arguments that returns a MapSet of alert IDs.

```ex
alert_fetcher: (() -> MapSet.t(SignsUi.Alerts.Alert.id())`
```

Update the `init` function to to get an `alert_fetcher` out of the options (which our tests will use), and use the real thing as a default.

```ex
alert_fetcher = Keyword.get(opts, :alert_fetcher, &SignsUi.Alerts.State.active_alert_ids/0)
```

Update the `init` return value to put this alert fetcher in the GenServer's state as it starts up:

```ex
{:ok, %{alert_fetcher: alert_fetcher, ....}}
```

So now that the alert fetcher is in the state, the GenServer can use it by: `state.alert_fetcher.()`.

## Expiring alerts with the alert data

Now that I've looked more closely at the GenServer and thought about it a little more, I think there's actually a better way than what I sent you before. Rather than doing two passes, first expiring by time, and then expiring by alert, it's probably easiest to just tap into the one pass.

So first, I'd recommend undoing the rename to `expire_signs_via_time`. Instead, `expire_signs` will do both kinds of expiration. That means it needs to take the `alert_fetcher` as an argument, just like it takes the `time_fetcher` now.

Then `expire_signs` can get the active alerts, just like it gets the current time:

```ex
def expire_signs(state, time_fetcher, alert_fetcher) do
  current_dt = time_fetcher.()
  alert_ids = alert_fetcher.()
  # ...
```

You would then update `expire_single_sign` to take three arguments, the last one being `alert_ids`, and add another clause to the function for when a sign config's `:alert_id` is not `nil`, similar to the clause that checks for `:expires`.

To test it, in `test/signs_ui/config/expiration_test.exs`, add a new `test` to the `describe "expire_signs/2"` block `test "produces correct updates when alerts expire"`. It should work roughly the same as the `"produces correct updates"` test (which you should rename to `"produces correct updates when times expire"`), but you can provide a sign state similar to the one in your test here, with signs in various modes with various alert IDs. And now, because the `expire_signs/3` function now takes a 3rd argument which is a function that returns the alerts, you can pass in `fn -> MapSet.new(your_alert_ids) end`.

lib/signs_ui/alerts/state.ex
```diff
@@ -24,6 +24,20 @@ defmodule SignsUi.Alerts.State do
     {:ok, %__MODULE__{}}
   end
 
+  @spec active_alert_ids(GenServer.server()) :: [Alert.id()]
```
losvedir, line 27: Actually, this function returns a `MapSet.t(Alert.id())`.

lib/signs_ui/alerts/state.ex
```diff
@@ -24,6 +24,20 @@ defmodule SignsUi.Alerts.State do
...
+  def active_alert_ids(pid \\ __MODULE__) do
+    GenServer.call(pid, :active_alert_ids)
+  end
+
+  def handle_call(:active_alert_ids, _from, state) do
+    alerts = state.alerts
+    routes = Map.keys(alerts)
+    alert_ids = for route <- routes, do: Map.keys(alerts[route])
```
losvedir, line 35: `for` expressions are quite powerful. You can deconstruct further by adding more expressions in more commas, and you can deconstruct a map into key-value pairs. So this, should get you a nice flat list of alert_ids in one go:

```ex
alert_ids =
  for {route_id, route_alerts} <- state.alerts,
      {alert_id, _alert_data} <- route_alerts,
      do: alert_id
```

lib/signs_ui/config/expiration.ex
```diff
@@ -47,22 +56,33 @@ defmodule SignsUi.Config.Expiration do
     {:noreply, state}
   end
 
-  @spec expire_signs(SignsUi.Config.State.t(), (() -> DateTime.t())) :: %{
```
losvedir, line 50: I think you accidentally removed this spec.

lib/signs_ui/config/expiration.ex
```diff
@@ -47,22 +56,33 @@ defmodule SignsUi.Config.Expiration do
...
     |> Map.get(:signs)
-    |> Enum.flat_map(&expire_single_sign(&1, current_dt))
+    |> Enum.flat_map(fn sign -> expire_single_sign(sign, current_dt, alert_ids) end)
     |> Enum.into(%{})
   end
 
-  @spec expire_single_sign({Sign.id(), Sign.t()}, DateTime.t()) :: [{Sign.id(), Sign.t()}]
+  @spec active_alert_ids(SignsUi.Alerts.State.state()) ::
+          MapSet.t(String.t())
+  def active_alert_ids(alert_state) do
```
losvedir, line 71: This is now in the other module.

lib/signs_ui/config/expiration.ex
```diff
@@ -72,7 +92,19 @@ defmodule SignsUi.Config.Expiration do
     end
   end
 
-  defp expire_single_sign(_, _) do
+  defp expire_single_sign(
+         {id, %Sign{config: %{alert_id: alert_id}} = sign},
+         _current_dt,
+         alert_ids
+       ) do
+    if is_nil(alert_id) or MapSet.member?(alert_ids, alert_id) do
```
losvedir, line 100: Can you please move the "not nil" check into the function header to be a guard, like in the `:expires` clause?

test/signs_ui/alerts/state_test.exs
```diff
@@ -12,6 +12,33 @@ defmodule SignsUi.Alerts.StateTest do
     end
   end
 
+  describe "handle_call :active_alert_ids" do
+    test "returns the alert_ids out of the state" do
+      state = %SignsUi.Alerts.State{
+        alerts: %{
+          # fill in sample data here, but it's of this shape:
```
losvedir, line 19: No need for this comment. I left it for you, to add more sample data, but actually I think these three alerts are fine.

test/signs_ui/config/expiration_test.exs
```diff
@@ -5,8 +5,8 @@ defmodule SignsUi.Config.ExpirationTest do
   alias SignsUi.Signs.Sign
   alias SignsUi.Config.Sign
 
-  describe "expire_signs/2" do
-    test "produces correct updates" do
+  describe "expire_signs_via_time/2" do
```
losvedir, line 8: The function wasn't renamed, in the end.

test/signs_ui/config/expiration_test.exs
```diff
@@ -50,9 +51,141 @@ defmodule SignsUi.Config.ExpirationTest do
...
+            "1234" => %{},
+            "5678" => %{}
+          },
+          "Blue" => %{
+            "abc" => %{}
+          }
+        }
+      }
+
+      alert_ids = SignsUi.Config.Expiration.active_alert_ids(alert_state)
```
losvedir, line 76: This is from before. You don't need this part of the test anymore since the code was moved to the other module.

test/signs_ui/config/expiration_test.exs
```diff
@@ -50,9 +51,141 @@ defmodule SignsUi.Config.ExpirationTest do
...
+          id: "471",
+          config: %{mode: :auto}
+        },
+        "273" => %Sign{
+          id: "273",
+          config: %{mode: :auto}
+        }
+      }
+
+      assert MapSet.new(expired_signs) == MapSet.new(expected_signs)
```
losvedir, line 188: No need to do `MapSet` here. We only did that for the other tests because the return value of the function actually was a MapSet. But here they're not. You should just be able to do `assert expired_signs == expected_signs`.

lib/signs_ui/config/expiration.ex
```diff
@@ -47,22 +56,26 @@ defmodule SignsUi.Config.Expiration do
     {:noreply, state}
   end
 
-  @spec expire_signs(SignsUi.Config.State.t(), (() -> DateTime.t())) :: %{
+  @spec expire_signs(SignsUi.Config.State.t(), (() -> DateTime.t()), (() -> String.t())) :: %{
```
losvedir, line 59: The third argument here, the `alert_fetcher` is a function that returns a mapset of alert IDs, so it should have a spec of `(() -> MapSet.t(SignsUi.Alerts.Alert.id()))`.

losvedir:
I left one small comment, but otherwise it looks good to me. Before merging, let's take it for a spin on signs-ui-dev and make sure nothing unexpected happens. Can you deploy your branch to signs-ui-dev, and then click around and make sure things are working as expected? And then search splunk for `index=signs-ui-dev warn OR error OR crash`.

mkennedm:
> I left one small comment, but otherwise it looks good to me. Before merging, let's take it for a spin on signs-ui-dev and make sure nothing unexpected happens. Can you deploy your branch to signs-ui-dev, and then click around and make sure things are working as expected? And then search splunk for `index=signs-ui-dev warn OR error OR crash`.

@losvedir 

https://signs-dev.mbtace.com/viewer looks fine, Splunk has this error every 5 seconds.

`22:10:13.263 [error] GenServer SignsUi.Config.Expiration terminating`

losvedir:
> but Splunk has this error every 5 seconds

If you click one of those and go to "Event actions" -> "Show source" it will show you the logs right there for more context. The error is:

```
** (stop) exited in: GenServer.call(SignsUi.Alerts.State, :active_alert_ids, 5000)
  ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started 
(elixir 1.11.4) lib/gen_server.ex:1017: GenServer.call/3
 (signs_ui 0.0.1) lib/signs_ui/config/expiration.ex:68: SignsUi.Config.Expiration.expire_signs/3
 (signs_ui 0.0.1) lib/signs_ui/config/expiration.ex:45: SignsUi.Config.Expiration.handle_info/2
 (stdlib 3.14.1) gen_server.erl:689: :gen_server.try_dispatch/4
 (stdlib 3.14.1) gen_server.erl:765: :gen_server.handle_msg/6
 (stdlib 3.14.1) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```

So what it looks like from "there's no process currently associated with the given name", is `SignsUi.Alerts.State` can't be found because we never gave it a name. To fix that, I think you can update that GenServer's `start_link` like such:

```
  def start_link(opts \\ []) do
    {start_opts, init_opts} = Keyword.split(opts, [:name])
    start_opts = Keyword.put_new(start_opts, :name, __MODULE__) # this line is new
    GenServer.start_link(__MODULE__, init_opts, start_opts)
  end
```

That way, if we don't pass in a name when starting the GenServer, it will use `__MODULE__` (which is an Elixir shortcut for the name of the current module, i.e., `SignsUi.Alerts.State` in this case), as the name. Which is what we need since the Expiration GenServer is trying to call out to it to get the alerts, using the name `SignsUi.Alerts.State`.

losvedir:
Ah, so that actually introduced some test errors. The problem is now that we're starting up the GenServer with a default name, if we don't provide one, any GenServer that starts up like that will be given the same name. And you can't start multiple GenServers with the same name.

So you'll need to update the `SignsUi.Alerts.StateTest` where we start up a GenServer (i.e. `SignsUi.Alerts.State.start_link(...)`, and give those a name now. You can pass in `name: :some_name` as an argument to the `start_link` to do that. The name itself doesn't matter too much, as long as its unique. I like to give a name based on the test that it's a part of.

losvedir (approved):
🎉 looks good to me and seems to be running okay on signs-ui-dev!

Please *squash* all these commits before merging (or use the GitHub squash and merge button if it will do it for you).

lib/signs_ui/config/expiration.ex
```diff
@@ -47,22 +56,30 @@ defmodule SignsUi.Config.Expiration do
...
+          SignsUi.Config.State.t(),
+          (() -> DateTime.t()),
+          (() -> MapSet.t(SignsUi.Alerts.Alert.id()))
+        ) :: %{
           Sign.id() => Sign.t()
         }
-  def expire_signs(state, time_fetcher) do
+  def expire_signs(state, time_fetcher, alert_fetcher) do
     current_dt = time_fetcher.()
+    alert_ids = alert_fetcher.()
```
paulswartz, line 68: no need to change, but for the future: it's probably easier to test this function if it takes `current_dt` and `alert_ids` rather than functions which return those values.

paulswartz (approved):

PR #512, opened 2021-05-06
Track subway alerts

bklebe:
#### Summary of changes
**Asana Ticket:** [🔀 Signs-UI keeps track of subway alerts](https://app.asana.com/0/584764604969369/1200160485026822/f)

The `SignsUi.Alerts.State` GenServer has been converted to a GenStage. This GenStage accepts events from ServerSentEventStage, which is configured to pull events from the alerts API endpoint. Updates to state are broadcast to the front-end application.

This feature diverges slightly from what was specified in the ticket. The data structure suggested for state is not normalized, and will contain duplicate alerts when one alert affects multiple routes, thus complicating adding, removing, and updating alerts. To simplify updates and lookups, the GenStage's state model instead uses the alert ID as the primary key, and stores a set of affected routes with each alert. In order to ensure that existing code written for this feature does not break as a result of this decision, the function `display_state/1` converts the normalized state model into the one specified in the ticket before it is broadcast to the signs channel.

I am open to changing how this works, either to send the normalized state model instead of the one from the ticket, or to use the state model in the ticket, but I strongly believe that keying alerts on their ID is the best way to store them in the GenStage.

#### Reviewer Checklist


losvedir:
Woohoo, I love it! I will dive in with more detail a little later today, but just a quick comment to get you started now.

I think your reworking of the state is fine. I don't actually mind the de-normalization, but I didn't really consider that the event updates will come in keyed by alert ID, so in that case it makes sense to store the alerts by ID, for easy lookup and update.

I'll say just glancing through the module, that it's kind of big now and does a lot. But broadly, those new things are "parse and handle the incoming events", and "transform into your 'display' format and broadcast out the state". I wonder if those could be separated out into a pair of pure modules, something like `Alerts.Events` and `Alerts.Display`? I'm not exactly sure the boundaries of those, but I have a gut feeling that a lot of the functions currently in `Alerts.State` would go in one or the other.

I think this would make testing easier, too, since the actual event handling and broadcasting would be done by `Alerts.State`. So the testing of the public functions of those other modules would be simple pure input/output transformations.

lib/signs_ui/alerts/display.ex
```diff
@@ -0,0 +1,64 @@
...
+        |> Map.new()
+    }
+  end
+
+  @spec expand_routes({Alert.id(), Alert.multi_route()}) :: [Alert.single_route()]
+  defp expand_routes({id, alert}) do
+    # Take a multi_route_alert(), maps over its affected_routes, and create a
+    # single_route_alert() for every affected route.
+    alert.affected_routes
+    |> Enum.map(fn route ->
```
losvedir, line 36: Ah, made me realize we don't have `credo` set up for signs_ui, but we generally don't pipe just once, instead preferring here `Enum.map(alert.affected_routes, fn route -> ...`.

lib/signs_ui/alerts/display.ex
```diff
@@ -0,0 +1,64 @@
...
+      alerts:
+        state
+        # generate a list of single_route_alerts from it, and flatten,
+        |> Stream.flat_map(&expand_routes/1)
+        # group them by route,
+        |> Enum.group_by(fn alert -> alert.route end)
+        # convert each group's list of alerts into a map,
+        |> Stream.map(&alert_map_for_route/1)
+        # and convert the list of alert groups into a map.
+        |> Map.new()
```
losvedir, line 27: I think what you have here is understandable and nice. But never one to pass up a good code golfing sesh, I might take advantage of `Map.new/2`, which takes a mapping function, the `& &1` anonymous function syntax. That would leave something like:

```ex
state
|> Stream.flat_map(&expand_routes/1)
|> Enum.group_by(& &1.route)
|> Map.new(&alert_map_for_route/1)

def alert_map_for_route({route, alerts}) do
  {route, Map.new(alerts, &{&1.id, &1})}
end
```

lib/signs_ui/alerts/events.ex
```diff
@@ -0,0 +1,48 @@
+defmodule SignsUi.Alerts.Events do
+  alias SignsUi.Alerts.Alert
+  alias SignsUi.Alerts.State
+
+  @spec parse(String.t() | list() | map()) :: [State.row()] | State.row()
```
losvedir, line 5: I know they say the Elixir philosophy is "let it crash", but in my experience that's not actually a great solution. Here, what I imagine is if alerts-ui produces some garbage alert for some reason, that could get signs-ui stuck in a crash loop.

Since the data is coming from outside this application's boundary, I think we need to be a little more careful with it, and have this top level function, return `{:ok, State.row()} | {:ok, [State.row()]} | :error`, and kind of be a little more precise with the error handling throughout.

A lot of the `{:ok, _} = ...` chains you have, can be made fallible with the `with` expression:

```ex
with 
  {:ok, created_at_local, _} <- DateTime.from_iso8601(attributes["created_at"]),
  {:ok, created_at_utc} <- DateTime.shift_zone,
  routes <- parse_routes(attributes)
do
  {:ok, {....}}
else
 _ -> :error
end
```

Although, come to think of it as I type that up, doesn't `DateTime.from_iso8601` already return the DateTime in UTC?

lib/signs_ui/alerts/events.ex
```diff
@@ -0,0 +1,48 @@
...
+       service_effect: service_effect,
+       affected_routes: routes
+     }}
+  end
+
+  @spec parse_attributes(nil | map()) ::
+          {DateTime.t() | nil, String.t() | nil, MapSet.t(Alert.route_id()) | nil}
+  defp parse_attributes(nil), do: {nil, nil, nil}
+
+  defp parse_attributes(attributes) do
```
losvedir, line 33: Looking at the definition of `Alert.t()`, we allow `created_at` and `service_effect` to both be `nil`, so if that's the case for either of them, we should still be able to get a valid `%Alert{}` out of the data.

test/signs_ui/alerts/state_test.exs
```diff
@@ -47,23 +50,163 @@ defmodule SignsUi.Alerts.StateTest do
...
 
-  # to be removed, once the real implementation is in:
-  describe "handle_info(:twiddle_state)" do
-    test "rolls randomly both code paths, and works without crashing" do
-      state =
-        Enum.reduce(0..20, %SignsUi.Alerts.State{}, fn _, state ->
-          {:noreply, new_state} = SignsUi.Alerts.State.handle_info(:twiddle_state, state)
-          new_state
-        end)
+  describe "handle_events" do
```
losvedir, line 53: I think this gets us to 100% code coverage for now, but with error handling we'll need a few more failure cases. I think it would be good to have a `display_test.exs` and an `alerts_test.exs`. We generally have a 1-to-1 mapping of code modules to test modules.

losvedir (commented):
On the whole looks great! The new `SignsUi.Alerts.State` is `:chefs_kiss`. I have some suggestions and tweaks, but I don't think anything too structurally major.

lib/signs_ui/alerts/display.ex
bklebe: Good point.

lib/signs_ui/alerts/events.ex
bklebe: That's a good point, I didn't realize ISO8601 doesn't contain offset information.

lib/signs_ui/alerts/alert.ex
```diff
@@ -9,4 +9,19 @@ defmodule SignsUi.Alerts.Alert do
           created_at: DateTime.t() | nil,
           service_effect: String.t() | nil
         }
+
+  @type route_id :: String.t()
+
+  @type multi_route :: %{
+          created_at: DateTime.t() | nil,
```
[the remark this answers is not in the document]
bklebe: The idea here was I wanted to make it impossible for the alert ID in the value to differ from the key, but your point is more pragmatic and makes more sense.

lib/signs_ui/alerts/events.ex
losvedir: Well, I think it can, but `from_iso8601` will shift it for you:

```
iex> {:ok, datetime, 9000} = DateTime.from_iso8601("2015-01-23T23:50:07,123+02:30")
iex> datetime
~U[2015-01-23 21:20:07.123Z]
```

lib/signs_ui/alerts/events.ex
bklebe: I'm not sure what case you're referencing here but if it's `parse_attributes(nil)`, that case only matches if the actual `attributes` field in the event is nil, which means we can't get either `created_at` or `service_effect` out of it.

lib/signs_ui/alerts/display.ex
```diff
@@ -0,0 +1,43 @@
...
+    |> Map.values()
+    |> Stream.flat_map(&expand_routes/1)
+    |> Enum.group_by(& &1.route)
+    |> Map.new(&alert_map_for_route/1)
+  end
+
+  @spec expand_routes(Alert.multi_route()) :: [Alert.t()]
+  defp expand_routes(alert) do
+    Enum.map(alert.affected_routes, fn route ->
+      %{
```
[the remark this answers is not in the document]
bklebe: Well, that would explain what I was saying about structs. Thanks for catching this.

lib/signs_ui/alerts/events.ex
```diff
@@ -0,0 +1,58 @@
...
+  def parse(payload) when is_list(payload) do
+    Enum.map(payload, &parse/1)
+  end
+
+  def parse(payload) when is_map(payload) do
+    id = payload["id"]
+
+    {created_at, service_effect, routes} = parse_attributes(payload["attributes"])
+
+    %{
```
losvedir, line 23: Returning a map rather than an `Alert` (honestly, Dialyzer, what would you even say you do here....? 😞 ).

lib/signs_ui/alerts/state.ex
```diff
@@ -4,79 +4,83 @@ defmodule SignsUi.Alerts.State do
...
   def init(opts) do
-    interval_ms = Keyword.get(opts, :interval_ms, 15_000)
-    :timer.send_interval(interval_ms, :twiddle_state)
-    {:ok, %__MODULE__{}}
+    {:consumer, %{}, opts}
   end
 
-  @spec active_alert_ids(GenServer.server()) :: MapSet.t(Alert.id())
+  @impl GenStage
+  def handle_info(msg, state) do
```
losvedir, line 29: Actually, I think you can just delete this. Basically, we need to provide a fallback `handle_info` to handle unexpected messages, if we define our own `handle_info` for certain messages, which I had done with `:twiddle_state`. However, now that `:twiddle_state` is gone, we can remove this fallback. If we don't define *any* `handle_info` in the module, the `use GenStage` macro provides a reasonable default like this for us.

lib/signs_ui/alerts/state.ex
```diff
@@ -4,79 +4,83 @@ defmodule SignsUi.Alerts.State do
...
-    Logger.info("new_alert_state: #{inspect(new_alert_state)}")
-
-    {:noreply, put_in(state.alerts, new_alert_state)}
+  @impl GenStage
+  @spec handle_events([Event.t()], GenStage.from(), t()) :: {:noreply, [], t()}
+  def handle_events(events, _from, state) do
+    # Works in two primary phases. First, we generate fresh t using the
+    # provided events:
+    new_state = update_state(events, state)
+    Logger.debug(["Alert state updated: ", inspect(new_state)])
```
losvedir, line 52: Can you make this an `info` actually, and provide a nice search key like `alert_state_updated`? As you mentioned, `server_sent_event_stage` already logs at the debug level, so we could use that potentially if needed, but the app generally doesn't run at debug. Since we only expect a handful of events here a minute, I think, it's fine to just log at info. I suspect we may peek in the logs periodically for this, without wanting to run at full debug logging.

test/signs_ui/alerts/display_test.exs
```diff
@@ -0,0 +1,67 @@
...
+  use ExUnit.Case, async: true
+  alias SignsUi.Alerts.Alert
+  alias SignsUi.Alerts.Display
+
+  describe "format_state/1" do
+    test "formats a state map" do
+      created_at = DateTime.new!(~D[2021-05-11], ~T[18:49:28], "Etc/UTC")
+
+      state = %{
+        "some_id" => %{
```
losvedir, line 11: I think these nested maps here should be `%Alert{}` structs, too?

lib/signs_ui/alerts/state.ex
bklebe: Sure.

.envrc.template
```diff
@@ -0,0 +1,2 @@
+export API_V3_HOST=
```
losvedir, line 1: I updated signs-ui-dev and signs-ui-prod to have their keys. However, as I was doing so, I realized that `_HOST` is not quite correct here, as that would be, e.g., `api-v3.mbta.com`. But the code expects both the host and the scheme. And we don't want to hardcode the scheme of `https` since then we can't swap in `http://localhost:xyz` as an environment variable.

Cora and I recently had a discussion about this on another PR and landed on the term `origin`, which is what the browser calls scheme+host, and I think that's probably the best term. I did a quick search now and can't really find anything better. Better than C#'s "left part" anyway.

So the ENV vars I put in AWS just now were `API_V3_ORIGIN` and `API_V3_KEY`, if you would please update the code here to follow.

lib/signs_ui/application.ex
```diff
@@ -16,7 +16,13 @@ defmodule SignsUi.Application do
       {SignsUi.Signs.State, [name: SignsUi.Signs.State]},
       SignsUi.Config.Expiration,
       SignsUi.RefreshTokenStore,
-      SignsUi.Alerts.State
+      {Application.get_env(:signs_ui, :alert_producer),
+       name: AlertProducer,
+       url: "#{System.get_env("API_V3_HOST")}/alerts?filter[datetime]=NOW&filter[route_type]=0,1",
```
losvedir, line 21: Gotta update here, too. ;) And the README, looks like.

bklebe: Wish we could have caught that with tests. Maybe this needs integration tests?

losvedir: Hm, testing ENV vars is hard. Even an integration test would probably run with `MIX_ENV=test`, I'd think. 🤔 

Our dev environment would have caught it, and we check the logs and click around there before deploying... in some sense that's like an integration test! I'd be open to a better way, but I can't think of how we could have caught this specific failure automatically.

losvedir (approved):
Looks good to me, once you run `mix format` one more time! 🎉 

Please squash all the commits down before merging, and use a conventional commit (`feat: ...`) for the commit message.

PR #525, opened 2021-05-19
Add Sentry to frontend

bklebe:
#### Summary of changes
**Asana Ticket:** [💳 Add sentry to signs-ui frontend](https://app.asana.com/0/584764604969369/1200158594431778/f)

I added Sentry's React SDK to the frontend app and configured it to load in the same environments as the backend Elixir SDK ("prod" and "dev"). I used a very similar pattern to Skate's implementation, where the global app template checks if it should configure Sentry and renders a script fragment that adds the DSN to `window` if it should. I had to loosen the CSP slightly to add the Sentry DSN to `connect-src`. I also removed an extra line in the Webpack config, and added some Webpack configuration for loading Sentry's source maps, because the browser console was throwing out a lot of warnings about not being able to load source maps from `node_modules/`.

#### Reviewer Checklist

### DIFF
--- a/lib/signs_ui_web/router.ex
+++ b/lib/signs_ui_web/router.ex
@@ -9,7 +9,7 @@ defmodule SignsUiWeb.Router do
 
     plug(:put_secure_browser_headers, %{
       "content-security-policy" =>
-        "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
+        "default-src 'self'; connect-src 'self' *.sentry.io; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
     })
   end
 
--- a/lib/signs_ui_web/router.ex
+++ b/lib/signs_ui_web/router.ex
@@ -9,7 +9,7 @@ defmodule SignsUiWeb.Router do
 
     plug(:put_secure_browser_headers, %{
       "content-security-policy" =>
-        "default-src 'self'; connect-src 'self' *.sentry.io; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
+        "default-src 'self'; connect-src 'self' #{Application.get_env(:sentry, :dsn, "")}; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
     })
   end
 
--- a/lib/signs_ui_web/router.ex
+++ b/lib/signs_ui_web/router.ex
@@ -9,7 +9,7 @@ defmodule SignsUiWeb.Router do
 
     plug(:put_secure_browser_headers, %{
       "content-security-policy" =>
-        "default-src 'self'; connect-src 'self' #{Application.get_env(:sentry, :dsn, "")}; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
+        "default-src 'self'; connect-src 'self' https://*.ingest.sentry.io; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
     })
   end
 
--- a/lib/signs_ui_web/templates/layout/app.html.eex
+++ b/lib/signs_ui_web/templates/layout/app.html.eex
@@ -10,8 +10,8 @@
     <title>MBTA Realtime Signs</title>
     <link rel="stylesheet" href="<%= static_path(@conn, "/vendor/bootstrap.min.css") %>">
     <link rel="stylesheet" href="<%= static_path(@conn, "/css/app.css") %>">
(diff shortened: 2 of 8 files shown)


lib/signs_ui_web/router.ex
```diff
@@ -9,7 +9,7 @@ defmodule SignsUiWeb.Router do
 
     plug(:put_secure_browser_headers, %{
       "content-security-policy" =>
-        "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
+        "default-src 'self'; connect-src 'self' #{Application.get_env(:sentry, :dsn, "")}; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'"
```
paulswartz, line 12: Does this mess anything up if the DSN isn't defined?

lib/signs_ui_web/views/layout_view.ex
```diff
@@ -1,3 +1,8 @@
 defmodule SignsUiWeb.LayoutView do
   use SignsUiWeb, :view
+  @spec record_sentry?() :: boolean()
+  def record_sentry?() do
+    sentry_env = Application.get_env(:sentry, :environment_name)
+    sentry_env == "dev" || sentry_env == "prod"
```
paulswartz, line 6: Is there a reason to check for the environment name instead of checking for whether the DSN is configured?

lib/signs_ui_web/templates/layout/app.html.eex
```diff
@@ -10,6 +10,9 @@
     <title>MBTA Realtime Signs</title>
     <link rel="stylesheet" href="<%= static_path(@conn, "/vendor/bootstrap.min.css") %>">
     <link rel="stylesheet" href="<%= static_path(@conn, "/css/app.css") %>">
+    <%= if record_sentry?() do %>
+      <%= render "_sentry.html", assigns %>
```
paulswartz, line 14: Do you need to pass the assigns? It doesn't appear they're used in the template.

lib/signs_ui_web/router.ex
bklebe: It shouldn't, that's why I added `""` as a default. Open to other suggestions here, I thought about extracting this into a function but that seemed too heavy.

lib/signs_ui_web/views/layout_view.ex
bklebe: Probably not.

lib/signs_ui_web/templates/layout/app.html.eex
bklebe: This was copied from Skate which also doesn't use them, I wasn't aware that you could render without passing assigns.

lib/signs_ui_web/router.ex
paulswartz: Oh, I see; `'self'` is a reference to the current host, and DSN is the host we're connecting to? And if there's no DSN, the line looks like `connect-src 'self'  ;` and the extra space doesn't affect anything?

lib/signs_ui_web/templates/layout/app.html.eex
paulswartz: There's a default of empty: https://hexdocs.pm/phoenix/1.5.9/Phoenix.Controller.html#render/1

lib/signs_ui_web/router.ex
bklebe: Yes.

lib/signs_ui_web/templates/layout/app.html.eex
bklebe: oh nice! thanks.

lib/signs_ui_web/router.ex
paulswartz: No, the extra space seems okay. It's okay for `connect-src` to have a full URL and not only a domain name?

paulswartz:
> @paulswartz, do you have opinions about having the Sentry script tag be in a separate template vs just inline in `app.html.eex`? It occurs to me it's quite small and we're not loading a bunch of other tools like Skate does.

@bklebe I don't have strong feelings: I think either is fine.

lib/signs_ui_web/router.ex
bklebe: Hmmm. well, it seemed to work with the full URL but looking at the docs it seems it's only supposed to be the host. the host is different per DSN though so we would need to either extract it from the variable or use a wildcard. consequences of using the wildcard are not extreme, it would just mean that connecting to any DSN would be allowed.

paulswartz: Maybe `*.ingest.sentry.io`? That seems like it would reduce the blast radius pretty far.

bklebe: That was my thought.

paulswartz (approved):
🍰

PR #530, opened 2021-05-26
Sign groups expire

bklebe:
#### Summary of changes
**Asana Ticket:** [🔀 Sign groups expire](https://app.asana.com/0/584764604969369/1200222628914559/f)

I added code to expire sign groups, including a new SignGroup struct and SignGroups module. The functions that handle expiring sign groups live in those modules, including `SignGroups.active/3` and `SignGroup.expired?/3`. I also corrected and unified some usages of typespecs in `SignsUi.Config`.

#### Reviewer Checklist

### DIFF
--- a/lib/signs_ui/config/sign_group.ex
+++ b/lib/signs_ui/config/sign_group.ex
@@ -3,6 +3,7 @@ defmodule SignsUi.Config.SignGroup do
   Represents a sign group.
   """
 
+  require Logger
   alias SignsUi.Alerts.Alert
   alias SignsUi.Config.Sign
 
@@ -31,5 +32,30 @@ defmodule SignsUi.Config.SignGroup do
   def expired?(%{alert_id: alert_id}, _current_time, alerts)
       when not is_nil(alert_id) do
     not MapSet.member?(alerts, alert_id)
+  end
+
+  @spec from_json(%{String.t() => any()}) :: t()
+  def from_json(map) do
+    %__MODULE__{
+      sign_ids: map["sign_ids"],
+      line1: map["line1"],
+      line2: map["line2"],
+      expires: expiration_from_iso8601(map["expires"]),
+      alert_id: map["alert_id"]
+    }
+  end
+
+  @spec expiration_from_iso8601(String.t() | nil) :: DateTime.t() | nil
+  defp expiration_from_iso8601(nil), do: nil
+
+  defp expiration_from_iso8601(expiration) do
+    case DateTime.from_iso8601(expiration) do
+      {:ok, dt, 0} ->
+        dt
+
+      _ ->
+        Logger.warning("Invalid sign group expiration time #{inspect(expiration)} received.")
+        nil
+    end
   end
 end
--- a/lib/signs_ui/config/sign_group.ex
+++ b/lib/signs_ui/config/sign_group.ex
@@ -1,0 +1,30 @@
+defmodule SignsUi.Config.SignGroup do
+  @moduledoc """
+  Represents a sign group.
+  """
+
+  alias SignsUi.Alerts.Alert
+  alias SignsUi.Config.Sign
+
+  @enforce_keys [:sign_ids, :line1, :line2, :expires, :alert_id]
+  @derive Jason.Encoder
+  defstruct @enforce_keys
+
+  @type t() :: %__MODULE__{
+          sign_ids: [Sign.id()],
+          line1: String.t() | nil,
+          line2: String.t() | nil,
+          expires: DateTime.t() | nil,
+          alert_id: Alert.id() | nil
+        }
+
+  def empty do
+    %__MODULE__{
+      sign_ids: [],
+      line1: nil,
+      line2: nil,
+      expires: nil,
+      alert_id: nil
+    }
+  end
+end
--- a/lib/signs_ui/config/sign_group.ex
+++ b/lib/signs_ui/config/sign_group.ex
@@ -8,7 +8,8 @@ defmodule SignsUi.Config.SignGroup do
   alias SignsUi.Config.Sign
 
   @derive Jason.Encoder
-  defstruct [:line1, :line2, :expires, :alert_id, sign_ids: []]
+  @enforce_keys [:route_id]
+  defstruct [:line1, :line2, :expires, :alert_id, :route_id, sign_ids: []]
 
   @type t() :: %__MODULE__{
           sign_ids: [Sign.id()],
@@ -38,6 +39,7 @@ defmodule SignsUi.Config.SignGroup do
   def from_json(map) do
     %__MODULE__{
       sign_ids: map["sign_ids"],
+      route_id: map["route_id"],
       line1: map["line1"],
       line2: map["line2"],
       expires: expiration_from_iso8601(map["expires"]),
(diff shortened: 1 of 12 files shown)


test/signs_ui/config/sign_group_test.exs
```diff
@@ -0,0 +1,84 @@
...
+  describe "expired/3" do
+    test "does not expire a sign group without an alert or expiration time" do
+      assert not SignGroup.expired?(
+               %SignGroup{},
+               DateTime.new!(~D[2021-05-21], ~T[17:31:00]),
+               MapSet.new(["active_alert"])
+             )
+    end
+
+    test "does not expire a sign group with an active alert id" do
```
mkennedm, line 14: You can also use `refute` instead of `assert not`. https://hexdocs.pm/ex_unit/1.12/ExUnit.Assertions.html#refute/1

mkennedm (approved):
Looks good. Made one small suggestion.

lib/signs_ui/config/sign_group.ex
```diff
@@ -0,0 +1,61 @@
...
+
+  def expired?(%{alert_id: alert_id}, _current_time, alerts)
+      when not is_nil(alert_id) do
+    not MapSet.member?(alerts, alert_id)
+  end
+
+  @spec from_json(%{String.t() => any()}) :: t()
+  def from_json(map) do
+    %__MODULE__{
+      sign_ids: map["sign_ids"],
```
losvedir, line 40: Maybe `Map.get(map, "sign_ids", [])`, so that per the typespec it's still a list if it's not present for some reason?

lib/signs_ui/config/sign_group.ex
```diff
@@ -0,0 +1,61 @@
...
+      alert_id: map["alert_id"]
+    }
+  end
+
+  @spec expiration_from_iso8601(String.t() | nil) :: DateTime.t() | nil
+  defp expiration_from_iso8601(nil), do: nil
+
+  defp expiration_from_iso8601(expiration) do
+    case DateTime.from_iso8601(expiration) do
+      {:ok, dt, 0} ->
```
losvedir, line 53: I think `_offset` might be better than `0` here. Even if there is an offset, the `dt` we get back is in UTC, so we can simply discard the offset. I think this value comes from the frontend, and its exact format might vary by browser or something. In any case, I don't think there's any reason we need to assert that there's no offset.

test/signs_ui/config/sign_group_test.exs
bklebe: Oh nice, thanks!

lib/signs_ui/config/state.ex
```diff
@@ -49,6 +48,11 @@ defmodule SignsUi.Config.State do
     GenServer.call(pid, {:update_chelsea_bridge_announcements, changes})
   end
 
+  @spec update_sign_groups(GenServer.server(), SignGroups.t()) :: {:ok, t()}
```
losvedir, line 51: Following up from our Slack thread:

None of the `update` functions are really documented, so can you add a little `@doc ""` to each of them?

In terms of how they work, there's two "kinds": replace state, and merge state.

The updates for Headways and Chelsea Bridge just take an entirely new state, and replace the old state with it. The Chelsea Bridge value is just a boolean... so that makes sense. For headways, I think it's probably not ideal, but that's how it is.

The `update_sign_config` function, in contrast, takes just *new updates* and takes care of merging into the state for you. This is good since a PIO might update a single sign or a few signs out of the hundreds, and so the frontend only needs to send those minimal values, and then on the backend the channel only needs to send the diff to the GenServer.

With GenServers, it's useful to think about what kind of data is being passed back and forth in the messages, and all else equal, keep it minimal if possible. So while `SignsUi.Config.State` knows the whole state, entities working with it might not have to.

In this branch, the expiration logic does have to fetch the entire sign group state from the `Config.State` GenServer, to know what's expired. However, it _doesn't_ necessarily need to send the whole thing back. In this case, we're not talking about a _ton_ of data, so it doesn't matter too much, but I think it's a good principle to follow.

And then when thinking about how the frontend will interact with the state here, _it_ will only be sending updates for a particular sign group at a time.

So I think it would be ideal if this was more of a "merge" function like with signs, and not a "replace" function like headways.

And one more thing, from that other ticket – I think in spec'ing out these tickets I expected that other one to be done first, so sorry for not mentioning it in this one – the frontend can _delete_ a sign group by sending the unix timestamp key and an empty map.

Can you try changing this function a bit, then, to merge data like `%{"Red" => %{"1234" => %{}, "5678" => %{....}}` into the state, in this case, deleting `"1234"` from the state, and adding `"5678"` to it? And then with that behavior, the expiration function would be building up a map of routes/timestamps of empty maps, signalling `Config.State` to delete them.

lib/signs_ui/config/sign_groups.ex
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  @spec from_json(t()) :: t()
+  def from_json(groups) do
+    Map.new(groups, fn {time, group} -> {time, SignGroup.from_json(group)} end)
+  end
+
+  @spec by_route(t()) :: by_route()
+  def by_route(groups) do
+    groups
```
losvedir, line 33: I think this is fine, but just as an option, it can be done in a single pass with a reduce, and `Map.update` which has an optional 4th argument if the key is already present:

```ex
Enum.reduce(sign_groups, %{}, fn {time, group}, acc ->
  Map.update(acc, group.route_id, %{time => group}, &(Map.put(&1, time, group)) )
end)
```

lib/signs_ui/config/sign_groups.ex
```diff
@@ -0,0 +1,50 @@
...
+    |> Enum.into(%{})
+  end
+
+  @spec update({unix_time(), SignGroup.t() | %{}}, t()) :: t()
+  def update({created_at, map}, groups) when map == %{}, do: Map.delete(groups, created_at)
+
+  def update({created_at, %SignGroup{sign_ids: new_sign_ids} = group}, groups) do
+    groups
+    |> Enum.map(fn {created_at, %SignGroup{sign_ids: sign_ids} = group} ->
+      {created_at, %{group | sign_ids: Enum.filter(sign_ids, &(&1 not in new_sign_ids))}}
```
losvedir, line 45: Rather than `filter` / `not in`, maybe `Enum.reject` / `in` instead? Feels a little double-negativey.

losvedir (approved):
Looks great, thanks! A few, small, optional comments, but otherwise :shipit: .

lib/signs_ui/config/sign_group.ex
```diff
@@ -0,0 +1,63 @@
...
+  Represents a sign group.
+  """
+
+  require Logger
+  alias SignsUi.Alerts.Alert
+  alias SignsUi.Config.Sign
+
+  @derive Jason.Encoder
+  @enforce_keys [:route_id]
+  defstruct [:line1, :line2, :expires, :alert_id, :route_id, sign_ids: []]
```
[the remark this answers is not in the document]
bklebe: good catch

lib/signs_ui/config/sign_groups.ex
bklebe: Just when I thought I had the most elegant solution!

lib/signs_ui/config/sign_groups.ex
bklebe: I always forget that `reject` is a thing!

mkennedm (approved):
Looks good.

PR #531, opened 2021-05-28
Add SignGroup form

losvedir:
#### Summary of changes
**Asana Ticket:** [🔀 Bulk create form](https://app.asana.com/0/281253957434160/1200222628914555)

This introduces the bulk edit form.

Screen Shot 2021-05-28 at 12 36 35 PM

It differs from [the mocks](https://app.abstract.com/projects/fd717ad5-2852-4575-ab6a-8adb1dc7aab8/branches/master/commits/latest/files/856b5444-1630-405c-a03b-1cef8145f675/layers/6FF1C212-14A6-459E-967F-2677998554FE?collectionId=7f32ea74-b391-49f2-93ef-f25f4453c8b3&collectionLayerId=45edca85-3394-4c5f-87ae-e69830bbf15b&mode=build&selected=root-BCD3D875-314E-46C9-BDEF-7582730C59C7) regarding the "Schedule return to Auto" section, because Kristin is working on a new version of how that will work.

There's minimal backend changes, just enough so I could make use of the callbacks and have state at the highest level of the frontend. The backend serializes the state to the frontend on initial page load (so this will have to change if we're changing how we're storing that state), and the channel GenServer just reflects the change right back to the frontend, to exercise the callbacks and see the data go to the backend, and then appear in the frontend. However, that frontend display of the created sign group is just a debug version. I realized we don't have an existing ticket for that part, so I made one for us to groom.

#### Reviewer Checklist


assets/css/sign_groups.css
```diff
@@ -0,0 +1,122 @@
...
+  justify-content: space-around;
+  background-color: white;
+  display: flex;
+  margin-bottom: 1rem;
+  padding: 1rem;
+}
+
+.sign_groups--virtual_sign {
+  /* Offset component's own spacing */
+  margin-left: -5rem;
```
digitalcora, line 98: Rather than doing something elsewhere and then undoing it here, can we apply the margin only where it's needed to begin with?

assets/css/virtual_sign.css
```diff
@@ -10,5 +10,5 @@
   height: 40px;
   padding-top: 8px;
   white-space: pre;
-  width: 24ch;
+  width: 25ch;
```
digitalcora, line 13: Did we make a mistake here originally and signs do in fact accommodate 25 characters?

assets/js/SignGroups.tsx
```diff
@@ -1,7 +1,361 @@
...
+  config: StationConfig;
+  zone: Zone;
+  line: string;
+  selectedSigns: { [sign: string]: boolean };
+  onSignChange: (signId: string, selected: boolean) => void;
+  readOnly: boolean;
+  kind: 'default' | 'named';
+}
+
+function ZoneSelector({
```
digitalcora, line 26: Can we keep things to one component per file? This is a very long file as it is, and it's hard to follow what might be relevant for each component.

assets/js/SignGroups.tsx
```diff
@@ -1,7 +1,361 @@
...
+  alerts,
+  setSignGroup,
+  readOnly,
+}: SignGroupsFormProps): JSX.Element | null {
+  const [selectedSigns, setSelectedSigns] = React.useState<{
+    [sign: string]: boolean;
+  }>({});
+  const [line1, setLine1] = React.useState('');
+  const [line2, setLine2] = React.useState('');
+  const [expireDT, setExpireDT] = React.useState<Date | null>(null);
```
digitalcora, line 97: `expireAt` might be a more conventional name for a datetime value; though I understood `DT` fine, so I don't know if it matters much.

assets/js/SignGroups.tsx
```diff
@@ -1,7 +1,361 @@
...
+  );
+
+  const handleSubmit = React.useCallback(
+    (event) => {
+      event.preventDefault();
+
+      /* eslint-disable @typescript-eslint/no-unused-vars -- the _id, and _isChecked */
+      const signIds = Object.entries(selectedSigns)
+        .filter(([_id, isChecked]) => isChecked)
+        .map(([id, _isChecked]) => id);
```
digitalcora, line 121: I think you can do this, and then we shouldn't need to disable the warning:

```js
.filter(([, isChecked]) => isChecked)
.map(([id, ]) => id);
```

assets/js/SignGroups.tsx
```diff
@@ -1,7 +1,361 @@
...
+                  onValidLine2Change={setLine2}
+                />
+              </div>
+              <div className="sign_groups--virtual_sign">
+                <VirtualSign line1={line1} line2={line2} time={currentTime} />
+              </div>
+            </div>
+          </div>
+          <div>
+            Schedule return to &quot;Auto&quot;
```
digitalcora, line 248: Is the `&quot;` actually needed in a text node? I don't think it would otherwise be a special character here.

assets/js/helpers.ts
```diff
@@ -87,4 +87,22 @@ function formatTime(dtIso8601: string | null): string {
   return `${datePart} @ ${timePart}`;
 }
 
-export { choosePage, alertsForLine, formatTime };
+function defaultZoneLabel(z: Zone): string {
+  /* eslint consistent-return: "off", default-case: "off" -- typescript ensures exhaustive match */
```
digitalcora, line 91: Are we already using a TypeScript profile for ESLint? Seems like it should include this, unless the exhaustive match is an optional feature.

assets/js/VirtualSign.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return `${text.padEnd(19)}${timeString(currentTime)}`;
+}
+
+interface VirtualSignProps {
+  time: number;
+  line1: string;
+  line2: string;
+}
+
+function VirtualSign({
```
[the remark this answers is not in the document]
losvedir: Ah, hm. I didn't intend "virtual" as a contrast to the signs lower on the page, but to try to capture the fact that it was a digital representation of the physical countdown signs in the stations. It's a little more than the sign content, there's the clock, too. Maybe just `Sign` works now that what used to be a `Sign` is now `SignPanel` to encompass the form widgets and such.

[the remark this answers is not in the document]
digitalcora: Alternative: `SignText`, to go along with `SignTextInput` (one displays the text that's on a sign, one lets you edit it).

assets/js/ViewerApp.tsx
```diff
@@ -170,6 +180,17 @@ class ViewerApp extends React.Component<
     }
   }
 
+  setSignGroup(ts: number, signGroup: SignGroup): void {
+    const { signGroupsChannel: channel } = this.state;
+
+    if (channel) {
```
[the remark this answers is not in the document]
losvedir: It's not present when the component's constructor is invoked and initializes the state. However, the channels are joined upon `componentDidMount`. So, I think it's correct for its type to be `Channel | null`, and to have to type-guard the null here, though in practice it should always be present by the time this function is invoked.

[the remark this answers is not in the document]
digitalcora: If it's something we expect should always be present (or else there's a bug in the code), maybe we can instead do `channel!.push(...)`? Then this would throw an exception, which would show up in Sentry, and we'd know it happened.

assets/css/virtual_sign.css
losvedir: In a sense... I suppose. I wanted to give a little "breathing room" to the right of the clock, since 24 characters is exactly the amount to the end of the clock. Since this is the `width`, I couldn't apply padding on that side (it eats into the line), and margin doesn't have the background color. I could rework it, but I thought just adding one more character of space was simple enough.

assets/js/SignGroups.tsx
losvedir: Yeah, I had that, and it does appease eslint, but it looked pretty subtle to me, easy to miss the comma, especially for the `map` where in a lot of situations trailing commas are optional. And I liked indicating that the first argument in the `filter` was `_id` to make the chained `map` a little easier to follow.

But, thinking about it now, I don't want to "write Elixir in Javascript", so since this is how it's done, I should probably just get with the program and get used to this pattern.

assets/js/SignGroups.tsx
losvedir: Yeah, otherwise runs afoul of `react/no-unescaped-entities`.

assets/css/virtual_sign.css
digitalcora: > Since this is the `width`, I couldn't apply padding on that side (it eats into the line)

It shouldn't work that way... is this element ending up with a non-standard `box-sizing`?

> By default in the CSS box model, the `width` and `height` you assign to an element is applied only to the element's content box. If the element has any border or padding, this is then added to the `width` and `height` to arrive at the size of the box that's rendered on the screen. [(source)](https://devdocs.io/css/box-sizing)

assets/js/helpers.ts
losvedir: I think we are. Frank switched us from tslint -> eslint, and our config seems to follow [the docs](https://github.com/typescript-eslint/typescript-eslint/blob/master/docs/getting-started/linting/README.md).

The `consistent-return` rule isn't *just* about types, strictly speaking. Some people don't seem to like the idea that a function can both early return and also just finish evaluating without returning anything, even if the type is `| undefined`. Similarly, `default-case` I guess is a style that some people like where a switch always has a `default`, even if it's empty.

But, personally, I don't think those provide much value on top of what TypeScript already gets us, and the `tsc` check in `npm run check` will highlight if a switch doesn't exhaustively handle cases, or if we have a code branch that doesn't return anything when it should.

So I suppose I'll remove the comment here and just disable those rules in the project instead.

assets/js/SignGroups.tsx
losvedir: Hmm, I'm ambivalent. My mental model is that this file is a module, and react components are just functions in a module. And so these non-exported ones are the equivalent of private functions, with all that entails regarding testability, freedom to change, etc. I think of `ZoneSelector` the same as if there was a little private helper function in here, like `shouldShowZone` or something. And I think of `SignGroups` as the single, shared component that the app uses.

I recognize that the file is pretty long, though. I feel like the react way of including markup sort of makes that unavoidable.

So what's the general principle we should have here? One react component per file? It doesn't matter to me that much, since I mostly just F12 to jump component to component. My only concern is that moving these private components to their own files means I have to export them, and then we'd test them, and then even though they're conceptually private we now have an implementation locked down. But since react components should _in principle_ be used anywhere, maybe that's okay. As part of this PR, for example, I did have to extract out two components that were previously private into their own public modules.

assets/css/virtual_sign.css
losvedir: Aha! Okay, good, I'm not crazy. I was going to respond to this comment, "I just did it this way without thinking much, but now realize I should use padding instead", but then when I added padding... it didn't work. I had a crisis of faith about my understanding of padding. So in that case, I think it must be overconstrained by a parent's width or some combination of flex shrink/grow. Let me fiddle with it a bit.

assets/js/VirtualSign.tsx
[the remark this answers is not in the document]
losvedir: Ah, I like `SignText` / `SignTextInput`.

assets/js/ViewerApp.tsx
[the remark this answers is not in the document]
losvedir: Interesting, yeah. Good idea.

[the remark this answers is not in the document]
losvedir: Hm, actually that runs afoul of a typescript lint we have configured, and I'm reluctant to disable that lint generally. However, I do like the idea of hearing in Sentry if there's an issue, so I'll use our JS Sentry package in an `else` block to send a message if it's not present for some reason.

assets/css/virtual_sign.css
losvedir: Actually, `24ch` is looking fine to me now, so I'll just change it back and leave it at that.

assets/js/ViewerApp.tsx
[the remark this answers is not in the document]
digitalcora: Equivalently to the `!` method, we could just `throw new Error("...")` and Sentry should pick that up without having to explicitly call it.

assets/css/virtual_sign.css
digitalcora: > I had a crisis of faith about my understanding of padding.

😂

assets/js/SignGroups.tsx
digitalcora: Hm, you're right that strict "one component per file" doesn't leave a lot of room for privacy of implementation details in the React world. I guess it's a balance we have to strike, like not having an excessive depth/complexity of private functions in an Elixir module. I think this is probably fine, my eyes just started to glaze over the first time I tried to read it, but it's more difficult to scan code on GitHub than in an editor.

digitalcora (approved):
🆎 🔗 ✏️ ✅

PR #552, opened 2021-07-09
Prompt if user tries to add a sign that's in another sign group to a new one

losvedir:
#### Summary of changes
**Asana Ticket:** [🔀 Handle signs that are in other sign groups](https://app.asana.com/0/584764604969369/1200222628914569)

First commit is a bonus that updates some Elixir deps.

Second commit introduces a `ModalPrompt` widget. I went back and forth on the implementation here, related to placement and the opacity.

One way to do it is to require the user of `ModalPrompt` to put it wherever you want it to cover. But thinking of it like a simple vanilla browser `prompt()` replacement, *that* would go in the widget dealing with the data, where the state is being updated. The only reason to put it higher is for CSS-ease, and making the `absolute` positioned opaque cover be able to do 100% width and height. You'd also have to go out of your way to put a `position: static` or `position: relative` on that container.

The second way, which I opted for, is to just have the `ModalPrompt` widget take an optional HTML tag ID, which it queries for its bounding box, and then have the `ModalPrompt` float an `absolute` positioned div with that same bounding box. It means optionally tagging a div with a certain ID, but that felt vaguely more robust than optionally tagging that div with a particular CSS property.

The third commit uses the `ModalPrompt` widget.

#### Reviewer Checklist


assets/css/modal_prompt.css
```diff
@@ -0,0 +1,16 @@
+.modal_prompt--container {
+  background-color: #d8d8d8;
+  border-radius: 4px;
+  box-shadow: 5px 5px 5px rgba(0, 0, 0, 0.2);
```
paulswartz, line 4: Not sure what the general principle is, but should this and `border-radius` be in `rem` or `em` instead of `px`?

assets/js/SignGroups.tsx
```diff
@@ -131,6 +172,23 @@ function SignGroupsForm({
 
   return (
     <form onSubmit={onSubmit}>
+      {changeGroupSignId && (
+        <ModalPrompt
+          acceptText="Yes, reassign this sign"
+          contents={changeGroupSignText(changeGroupSignId, line)}
+          elementId="tab-panel-container"
+          onAccept={() => {
+            const [station, zone] = changeGroupSignId;
```
paulswartz, line 181: is there a reason that this is `[station, zone]` instead of being `signId`?

assets/css/modal_prompt.css
losvedir: I think I got these from the Abstract mock-ups "inspect" section. But, yeah, since we try to size things in `rem` generally, I'll change it to that.

assets/js/SignGroups.tsx
losvedir: Oh, you're right the naming is kind of inconsistent. But I should probably go the other way, and not call it `changeGroupSignId`.

Basically, I started with the idea that we were storing the sign ID in the state here (e.g. `orange_north_station_commuter_rail_exit`), hence the name of the hooks, and what is now `[station, zone]` used to be `signId` like you suggest. However, I ran into issues because there's not a great way to get from `orange_north_station_commuter_rail_exit` to `North Station CR Exit`. The station name and zone/platform description are encoded in our configuration indexed by the ARINC identifiers (see [here](https://github.com/mbta/signs_ui/blob/56795cedbc561e60bd4248cdaca0608506ce1da7/assets/js/mbta.ts#L283)). We also can go ARINC to internal ID [here](https://github.com/mbta/signs_ui/blob/56795cedbc561e60bd4248cdaca0608506ce1da7/assets/js/mbta.ts#L2922) (in order to handle Realtime Signs POSTs).

I considered sticking with the ID, which would entail inverting that second map, and then string splitting on the `-` to get the ARINC station and zone, and then use those to index into the configuration. Or build another view of that configuration from the internal IDs. But ultimately, I realized that the component that's actually pressing the button is dealing with the ARINC station and zone *already*, and it was the component turning it into the internal ID, so instead I just had that component pass the station and zone along directly.

paulswartz: Thank you; I was looking in this function for the use of `station` and `zone` outside `signId`, but I see now that they're used separately in `changeGroupSignText`.

paulswartz (approved):

PR #554, opened 2021-07-13
fix: Updating SignGroups propagates to the sign configs

losvedir:
#### Summary of changes
**Asana Ticket:** [[extra] Saving sign groups updates the underlying signs' data.](https://app.asana.com/0/584764604969369/1200580099553918)

This adds a step in the `Config.State` handling of sign groups to reflect those changes across the individual signs as well.

#### Reviewer Checklist

### DIFF
--- a/lib/signs_ui/config/sign_group_to_sign_configs.ex
+++ b/lib/signs_ui/config/sign_group_to_sign_configs.ex
@@ -1,0 +1,116 @@
+defmodule SignsUi.Config.SignGroupToSignConfigs do
+  @moduledoc """
+  A compatibility layer that supports the concept of Sign Groups while leaving clients like Realtime
+  Signs unchanged. When Sign Groups are changed, those changes are propagated to the underlying Sign
+  configs. There are three kinds of changes:
+
+  * New Sign Group - all the signs in this group have their custom text, mode and expiration options
+    set to the same values as the group value.
+  * Sign Group Deleted - all the signs in the group are reset back to Auto
+  * Sign Group Updated - Same as New Sign Group, plus any signs that were removed are reset back to
+    Auto.
+  """
+
+  alias SignsUi.Config.Sign
+  alias SignsUi.Config.SignGroup
+  alias SignsUi.Config.SignGroups
+  alias SignsUi.Config.State
+
+  @doc """
+  Transforms the Sign Group changes into a map of Sign Config updates.
+  """
+  @spec apply(SignGroups.t(), State.t()) :: %{Sign.id() => Sign.t()}
+  def apply(sign_group_changes, config_state) do
+    Enum.reduce(sign_group_changes, %{}, fn {changed_sign_group_id, changed_sign_group}, acc ->
+      config_state.sign_groups
+      |> Map.get(changed_sign_group_id)
+      |> process_sign_group(normalize(changed_sign_group))
+      |> Map.merge(acc)
+    end)
+  end
+
+  @spec process_sign_group(SignGroup.t() | nil, SignGroup.t() | nil) :: %{Sign.id() => Sign.t()}
+  defp process_sign_group(old_sign_group, new_sign_group)
+
+  # Group is being deleted
+  defp process_sign_group(%SignGroup{} = old_group, nil) do
+    set_signs_to_auto(old_group.sign_ids)
+  end
+
+  # Group is being added
+  defp process_sign_group(nil, %SignGroup{} = new_sign_group) do
+    set_signs_to_static_text(
+      new_sign_group.sign_ids,
+      new_sign_group.line1,
+      new_sign_group.line2,
+      new_sign_group.expires,
+      new_sign_group.alert_id
+    )
+  end
+
+  # Group is being updated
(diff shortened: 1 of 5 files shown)


lib/signs_ui/config/sign_group_to_sign_configs.ex
```diff
@@ -0,0 +1,104 @@
...
+            expires: expires,
+            alert_id: alert_id
+          }
+        }
+      }
+    end)
+  end
+
+  defp normalize(%SignGroup{} = sign_group), do: sign_group
+  defp normalize(sg) when sg == %{}, do: nil
```
digitalcora, line 103: Any reason not to just put the `when sign_group == %{}` guard in the `process_sign_group` clauses themselves? Then you could ditch this and maybe also the "shouldn't occur" clause, and simplify the typespec.

lib/signs_ui/config/sign_group_to_sign_configs.ex
```diff
@@ -0,0 +1,104 @@
...
+  @doc """
+  Transforms the Sign Group changes into a map of Sign Config updates.
+  """
+  @spec apply(SignGroups.t(), State.t()) :: %{Sign.id() => Sign.t()}
+  def apply(sign_group_changes, config_state) do
+    Enum.reduce(sign_group_changes, %{}, fn {sign_group_id, sign_group}, acc ->
+      Map.merge(
+        process_sign_group(config_state.sign_groups[sign_group_id], normalize(sign_group)),
+        acc
+      )
```
digitalcora, line 28: Might be more legible as a pipeline:

```elixir
config_state.sign_groups[sign_group_id]
|> process_sign_group(sign_group)
|> Map.merge(acc)
```

digitalcora:
@losvedir Did you confirm that with these changes, using the "Ungroup" button on an individual sign returns it to auto? If so it'd be nice if we could fill in the `TODO` in `Line.test.ts` (adding an assertion that the sign has returned to auto).

lib/signs_ui/config/sign_group_to_sign_configs.ex
```diff
@@ -0,0 +1,104 @@
...
+
+    Map.merge(additions, removals)
+  end
+
+  # Shouldn't occur, but for completeness
+  defp process_sign_group(nil, nil) do
+    %{}
+  end
+
+  defp set_signs_to_auto(sign_ids) do
```
bklebe, line 78: Do we not do typespecs for private functions?

lib/signs_ui/config/sign_group_to_sign_configs.ex
```diff
@@ -0,0 +1,104 @@
...
+      {
+        sign_id,
+        %Sign{
+          id: sign_id,
+          config: %{
+            mode: :static_text,
+            line1: line1,
+            line2: line2,
+            expires: expires,
+            alert_id: alert_id
```
bklebe, line 95: What happens when a user tries to change the expiration on a sign that is currently grouped? It seems like from this it would just change. Do we need a way for a sign to know it's grouped (possible a different mode?), or updated logic to check when changing the sign?

bklebe (commented):
I agree with Cora's comments and have a couple questions of my own. Looks good overall, though.

lib/signs_ui/config/sign_group_to_sign_configs.ex
losvedir: I had it that way initially, but I felt that `%SignGroup{} | nil` (i.e.: `Option<SignGroup>`) was a better way of representing what the data was. The empty struct isn't semantically an empty struct there so much as a kind of flag.

We'd still need the "shouldn't occur" clause since `%{}` is not a subtype of `%SignGroup{}` and the pattern match fails.

lib/signs_ui/config/sign_group_to_sign_configs.ex
losvedir: We do, I just missed these.

lib/signs_ui/config/sign_group_to_sign_configs.ex
losvedir: It would, yeah. There's a separate task that prevents it on the front end. I could see an argument to also enforce it on the backend, but I don't think it's so critical that we need to worry about messing with the sign update logic. Basically, the configuration of each individual sign is the ultimate source of truth, and sign groups are merely a convenience for updating many of them at one time.

bklebe (approved):
Looks good! 🚀

losvedir:
> Did you confirm that with these changes, using the "Ungroup" button on an individual sign returns it to auto? If so it'd be nice if we could fill in the TODO in Line.test.ts (adding an assertion that the sign has returned to auto).

It does indeed! The branch is on signs-ui-dev now, if you want to try it out.

I looked at the TODO, but I'm unsure how exactly to write that part of the test. The functionality to propagate sign group changes to individual signs is all on the backend. Did you have something in mind there?

digitalcora:
> The functionality to propagate sign group changes to individual signs is all on the backend.

Oh, right. I was thinking (wishful thinking?) in terms of acceptance tests, but these are not those. Maybe we can just delete the `TODO` then, as it's not possible to implement such an assertion in that context.

PR #587, opened 2021-08-18
Reworking "Apply" in SignPanel

losvedir:
#### Summary of changes
**Asana Ticket:** [🔀 "apply" for both custom text modules functions similarly](https://app.asana.com/0/584764604969369/1200421875164011)

The first commit is a refactor of `SignPanel` to the functional component react style that we use now.

The second changes the component to not immediately save in all the cases that it used to. Formerly, the UX was a bit confusing: Changing to "Custom" immediately saved a new configuration with empty lines, having the effect of blanking the sign. Then subsequent changes to the line inputs didn't take effect until the "Apply" button was pressed. However, the expiration values (datetime or alert ID) were saved immediately upon choosing them. Now, changing to "Custom" exposes the form but does not save, and similarly adding an expiration criterion does not save. Only once "Apply" is pressed does it save everything. Note that per the ticket specification, changing to a mode other than Custom still *does* apply immediately.

To change the behavior in the desired way, I added more state to the component, and updated the things that previously saved to the backend, to simply update the internal state instead. And then "Apply" would send the new state to the backend. Since we're now keeping the whole sign config in the state, I removed the individual `staticLine1` and `staticLine2` fields, and replaced them with a `signConfig` object, which gets initialized by the current value in its props.

#### Reviewer Checklist

### DIFF
--- a/assets/js/SignPanel.tsx
+++ b/assets/js/SignPanel.tsx
@@ -1,12 +1,6 @@
 import * as React from 'react';
 import lineToColor from './colors';
-import {
-  RouteAlerts,
-  SignConfig,
-  SignConfigs,
-  SignGroup,
-  SingleSignContent,
-} from './types';
+import { RouteAlerts, SignConfig, SignGroup, SingleSignContent } from './types';
 import { choosePage } from './helpers';
 import SignGroupExpirationDetails from './SignGroupExpirationDetails';
 import SignTextInput from './SignTextInput';
@@ -106,29 +100,6 @@ function stringify(expires: Date | null) {
   return null;
 }
 
-function updateConfig(
-  setConfigsFn: (x: SignConfigs) => void,
-  realtimeId: string,
-  signConfig: SignConfig,
-  expires: Date | [Date, Date] | null,
-  alertId: string | null,
-) {
-  if (Array.isArray(expires)) {
-    return;
-  }
-  const expirationConfig = stringify(expires);
-
-  const newConfig = {
-    ...signConfig,
-    expires: expirationConfig,
-    alert_id: alertId,
-  };
-
-  setConfigsFn({
-    [realtimeId]: newConfig,
-  });
-}
-
 function parseDate(str: string | null | undefined): Date | null {
   if (str) {
     const date = new Date(str);
@@ -153,7 +124,7 @@ interface SignPanelProps {
   signContent: SingleSignContent;
   currentTime: number;
   line: string;
-  setConfigs: (x: SignConfigs) => void;
+  setConfig: (x: SignConfig) => void;
   signConfig: SignConfig;
   realtimeId: string;
   signGroup?: SignGroup;
@@ -173,27 +144,48 @@ function SignPanel({
   signContent,
   currentTime,
   line,
-  setConfigs,
-  signConfig,
+  setConfig,
+  signConfig: initialSignConfig,
   realtimeId,
   signGroup,
   ungroupSign,
   readOnly,
 }: SignPanelProps): JSX.Element {
-  const [staticLine1, setStaticLine1] = React.useState(signConfig.line1 || '');
-  const [staticLine2, setStaticLine2] = React.useState(signConfig.line2 || '');
-  const [customChanges, setCustomChanges] = React.useState(false);
+  const [signConfig, setSignConfig] = React.useState(initialSignConfig);
+  const [hasCustomChanges, setHasCustomChanges] = React.useState(false);
   const [initialTime] = React.useState(currentTime);
   const [confirmingUngroup, setConfirmingUngroup] = React.useState(false);
 
+  React.useEffect(() => {
+    if (!hasCustomChanges) {
+      setSignConfig(initialSignConfig);
+    }
+  }, [initialSignConfig, hasCustomChanges]);
+
+  function save(config: SignConfig): void {
+    setHasCustomChanges(false);
+    setConfig(config);
+  }
+
   function handleInputLine1(newText: string): void {
-    setStaticLine1(newText);
-    setCustomChanges(true);
+    setSignConfig({ ...signConfig, line1: newText });
+    setHasCustomChanges(true);
   }
 
   function handleInputLine2(newText: string): void {
-    setStaticLine2(newText);
-    setCustomChanges(true);
+    setSignConfig({ ...signConfig, line2: newText });
+    setHasCustomChanges(true);
+  }
+
+  function handleModeSelect(event: React.ChangeEvent<HTMLSelectElement>): void {
+    const value = event.target.value as SignModeOptions;
+    const newConfig = makeConfig(value);
+    setSignConfig(newConfig);
+    setHasCustomChanges(true);
+
+    if (value !== 'static_text') {
+      save(newConfig);
+    }
   }
 
   function performUngroup(): void {
@@ -201,21 +193,6 @@ function SignPanel({
       setConfirmingUngroup(false);
       ungroupSign();
     }
-  }
-
-  function saveStaticText(): void {
-    setCustomChanges(false);
-
-    const newConfig: SignConfig = {
-      ...signConfig,
-      mode: 'static_text',
-      line1: staticLine1,
-      line2: staticLine2,
-    };
-
-    setConfigs({
-      [realtimeId]: newConfig,
-    });
   }
 
   return (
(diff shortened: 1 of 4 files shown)


assets/js/SignPanel.tsx
```diff
@@ -161,58 +161,50 @@ interface SignPanelProps {
...
+  setConfigs,
+  signConfig,
+  realtimeId,
+  signGroup,
+  ungroupSign,
+  readOnly,
+}: SignPanelProps): JSX.Element {
+  const [staticLine1, setStaticLine1] = React.useState(signConfig.line1 || '');
+  const [staticLine2, setStaticLine2] = React.useState(signConfig.line2 || '');
+  const [customChanges, setCustomChanges] = React.useState(false);
```
digitalcora, line 161: This might be better named `hasCustomChanges`, else it sort of sounds like it _is_ the custom changes.

assets/js/SignPanel.tsx
```diff
@@ -161,58 +161,50 @@ interface SignPanelProps {
...
+  const [initialTime] = React.useState(currentTime);
+  const [confirmingUngroup, setConfirmingUngroup] = React.useState(false);
+
+  function handleInputLine1(newText: string): void {
```
digitalcora, line 169: Shouldn't these be defined with `useCallback`? (avoids re-defining them on every render if nothing has changed)

assets/js/SignPanel.tsx
```diff
@@ -226,199 +218,162 @@ class SignPanel extends React.Component<
...
+                id={realtimeId}
+                className="viewer--mode-select"
+                value={signConfig.mode}
+                onChange={(event) => {
+                  const newConfig = makeConfig(
+                    event.target.value as
+                      | 'auto'
+                      | 'headway'
+                      | 'off'
+                      | 'static_text',
```
digitalcora: Do we not have a named type defined for this?

assets/js/SignPanel.tsx
```diff
@@ -226,199 +218,162 @@ class SignPanel extends React.Component<
...
+              {displayName(signConfig.mode)}
+            </div>
+          )}
+          {!readOnly && !signGroup && (
+            <div>
+              <select
+                id={realtimeId}
+                className="viewer--mode-select"
+                value={signConfig.mode}
+                onChange={(event) => {
```
digitalcora, line 211: This function seems complex enough that maybe it shouldn't be defined inline in the markup. (at least, if something like `handleInputLine1` gets its own function, this should too)

assets/js/SignPanel.tsx
```diff
@@ -174,25 +151,28 @@ function SignPanel({
...
-  const [staticLine1, setStaticLine1] = React.useState(signConfig.line1 || '');
-  const [staticLine2, setStaticLine2] = React.useState(signConfig.line2 || '');
+  const [signConfig, setSignConfig] = React.useState(initialSignConfig);
   const [customChanges, setCustomChanges] = React.useState(false);
   const [initialTime] = React.useState(currentTime);
   const [confirmingUngroup, setConfirmingUngroup] = React.useState(false);
 
+  React.useEffect(() => {
+    setSignConfig(initialSignConfig);
+  }, [setSignConfig, initialSignConfig]);
```
digitalcora, line 167: I'm not sure of the reasoning here but it looks like this will set the sign config back to the "initial" (passed-in) config whenever that initial config changes. If you're in the middle of changing a sign's config, and it changes on its own due to a server update, wouldn't this wipe out your changes?

assets/js/SignPanel.tsx
```diff
@@ -203,19 +183,9 @@ function SignPanel({
...
-      ...signConfig,
-      mode: 'static_text',
-      line1: staticLine1,
-      line2: staticLine2,
-    };
-
-    setConfigs({
-      [realtimeId]: newConfig,
-    });
+    setConfigs({ [realtimeId]: config });
```
digitalcora, line 188: Just as the sign is passed only its own config (not all of them), could we pass each sign an update function that only updates its own config? Then individual signs wouldn't have to "know" the data model of the larger config-for-all-signs.

assets/js/SignPanel.tsx
losvedir: Mmm, it felt like a close call to me. Whether I do:

```typescript
function handleInputLine1(newText: string): void { ... }
```

or

```typescript
const handleInputLine1 = React.useCallback(() => { .... }, [...])
```

I'm creating/allocating a new function every render either way, right? (The first is the named function, the second is the anonymous arrow function literal).

I think of the main benefit of `useCallback` as referential transparency, so that when it gets passed as a prop to a child, the *child* doesn't have to re-render. But that depends on the size of the child, and in this case it was a pretty small little component, so it didn't feel like a big win. And using `React.useCallback` makes the initial render of *this* function a little slower.

In general, I don't have a great sense of where the tradeoff is, but it doesn't make sense to use `React.useCallback` as a general rule always, and when it's a toss up I prefer the simplicity of a normal definition.

Or am I missing something about the performance trade-offs here?

digitalcora: > it doesn't make sense to use `React.useCallback` as a general rule always

"Use it as a general rule always" was the impression I had, though I can't point to any official documentation that says it. If that's the wrong impression, I guess you already know more than me!

assets/js/SignPanel.tsx
losvedir: The thing that prompted it was needing the sign panel to update when the "All to Auto/Off/Headway" was pressed. Since the function won't re-render when the props change automatically - only the state - you would click "All to Auto" and the signs wouldn't respond. So basically, yeah, by design it blows away the state when the server and props change, and it would wipe out changes if someone was doing it at the same time, but I couldn't think of a simple way to handle that scenario, and it seemed pretty unlikely to me.

digitalcora: If we're sure this will only happen when someone else is editing _the same sign_ at the same time (and not just editing any sign), I'd agree it would be uncommon enough.

assets/js/SignPanel.tsx
losvedir: Pushed this fix up in a separate fixup commit because I'm kind of iffy on it now. It's certainly much cleaner, but now there's an explosion of memoized new functions. (In contrast to the individual config, where each `SignPanel` is just passed its own pointer into the same larger structure.) Still, these are sort of micro-optimizations so might be worth it for the improved clarity.

assets/js/SignPanel.tsx
losvedir: Yeah, that was my impression after going down a rabbit hole on it a while ago, but I can't reconstruct all the things that led me to that conclusion. I think [this blog post](https://kentcdodds.com/blog/usememo-and-usecallback) was one that made a big impression. I could have sworn I saw vjeux (big react guy at facebook) say something in agreement but now I can't find that.

I do recall hearing that advice for class components about always putting logic in methods, since those really will only be created once, but even then I think it was controversial.

assets/js/SignPanel.tsx
losvedir: Ack, good catch! `initialSignConfig` changes at a reference level if any sign is edited. I pushed up a commit that wraps the `setSignConfig(...)` call in `if (!hasCustomChanges)`, so if the form is open / there's unsaved changes, they won't get blown away.

digitalcora: That's probably a decent compromise. Though, if I push "All to X", does that reset any custom changes I made in my own browser? Because in that case I might expect it to... but I guess there are edge cases everywhere in the current "concurrency" model and we can't really catch them all without a larger refactoring.

digitalcora (approved):
Looks good from here!

assets/js/SignPanel.tsx
losvedir: Heh, no it leaves the form alone. But playing around with it, I think that feels fine. Say the sign is Off, and you open up the "Custom form", and someone clicks "All to Auto", that sign's config will change, and so the user will start to see the realtime-predictions on the preview of the sign. Meanwhile, their form will still be open, with their changes present and the "Apply" button enabled with the little "there are changes asterisk". In other words, it will be in exactly the state it would be in if someone had done "All to Auto" *first* and *then* someone went in and started changing the custom text. At that point, they can either cancel, and the sign will go back to Auto, or click Apply, and their changes will take effect. I think it works out pretty well.

PR #699, opened 2022-01-10
Add new glx stations to Signs UI

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Updates to signs ui](https://app.asana.com/0/1199504266721641/1200959142579183/f)

This PR adds options to the Signs UI for the GLX stations that are planned to be opened in the next couple months. This is a precursor to updates to realtime signs as well as field testing of these signs that were added.

Screen Shot 2022-01-24 at 11 21 23 AM

#### Reviewer Checklist


losvedir (commented):
I think you'll also want to update `arincToRealtimeIdMap` in `mbta.ts`, which maps the ARINC id/zone to some arbitrarily determined ID used internally by realtime_signs.

You might also want to enable `custom: true` on the signs, since putting some custom text on there in dev is one of the common ways that we test that realtime-signs and signs-ui are correctly integrated. Having two options also allows the frontend to toggle between the two, and fiddling with the values is what should actually end up persisting these new signs to the config JSON in S3.

PaulJKim:
That makes sense for adding the custom option.

So these mappings correspond to the ids in `priv/signs.json` over on Realtime Signs, right? After we set the id in the `mbta.ts` file here, we just need to configure Realtime signs to have those id's in that config file?

assets/js/mbta.ts
```diff
@@ -2761,13 +2806,27 @@ const stationConfig: {
...
+        },
+      },
       {
         id: 'GLEC',
         name: 'Lechmere',
         zonePositions: {
           left: ['n'],
           center: [],
-          right: ['m'],
+          right: ['s'],
```
PaulJKim, line 2829: Updated this one because there was an Id colliding with the new GLEC-m zone on the new train stop, and to more accurately map s to south.

losvedir:
> So these mappings correspond to the ids in `priv/signs.json` over on Realtime Signs, right? After we set the id in the `mbta.ts` file here, we just need to configure Realtime signs to have those id's in that config file?

Yep, exactly.

assets/js/mbta.ts
```diff
@@ -2761,13 +2806,27 @@ const stationConfig: {
...
+              auto: false,
+              custom: true,
+              headway: false,
+              off: true,
+            },
+          },
+        },
+      },
       {
         id: 'GLEC',
```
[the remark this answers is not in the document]
PaulJKim: Will definitely double check with Andrew on that. I figured that the `GLEC` zones belonged to the new train stop and the `SLEC` zone belonged to bus, but I wasn't sure if we wanted to still keep the old busway signs considering there are currently still buses running out of the old Lechmere station.

I guess that mainly begs the question of what exactly the rollout plan is for the new zones. Although these changes don't need to go to prod until the station officially opens, I wonder if, when that happens, it will be an immediate and full cutover to the new train station/busway or would there be any reason to continue running buses out of old lechmere for some time when the new station opens?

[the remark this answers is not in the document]
losvedir: Oh, I was under the impression the PA/ESS signs at the old Lechmere were offline. Can you check with Dan Salomon about that? And I think the new signs are up and running. If that's the case, I think you can do a hard cutover now, and we can just set them to "Off".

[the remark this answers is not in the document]
PaulJKim: Ah I see, I spoke to Dan and the signs are still powered on but because they're cut off from the network, there's nothing syncing them to keep them from drifting so I think the times are off by several hours. So it does seem like it'll be fine to completely remove them. Probably also will be worth reaching out to New Era Tech to have those signs turned off since they're not doing anything helpful.

assets/js/mbta.ts
```diff
@@ -1987,16 +1987,65 @@ const stationConfig: {
   Green: {
     batchModes: { auto: true, headway: true, off: true },
     stations: [
+      {
+        id: 'GUNS',
+        name: 'Union Sq',
+        zones: {
+          e: {
+            label: "North track",
```
PaulJKim, line 1995: @losvedir I added some labeling to GUNS and GLEC to help reflect the reality of the stations and with the aim of helping avoid confusion for users regarding the weirdness with cardinal directions along the GLX. Let me know if you have any thoughts or if you think this makes sense!

assets/js/mbta.ts
```diff
@@ -3135,15 +3169,19 @@ const arincToRealtimeIdMap: { [key: string]: string } = {
   'GSCI-e': 'science_park_eastbound',
   'GSCI-w': 'science_park_westbound',
   'GSCI-m': 'science_park_mezzanine',
-  'GLEC-w': 'lechmere_green_line',
+  'GLEC-e': 'lechmere_green_line_eastbound',
+  'GLEC-w': 'lechmere_green_line_westbound',
+  'GLEC-m': 'lechmere_green_line_mezzanine',
+  'GUNS-e': 'union_sq_north',
+  'GUNS-w': 'union_sq_south',
```
digitalcora, line 3176: Doesn't this have to match the realtime_signs config? So either this should be `east` and `west`, or realtime_signs should be `north` and `south`.

assets/js/mbta.ts
```diff
@@ -2762,27 +2811,12 @@ const stationConfig: {
...
       {
-        id: 'GLEC',
-        name: 'Lechmere',
-        zonePositions: {
-          left: ['n'],
-          center: [],
-          right: ['m'],
-        },
+        id: 'SLEC',
+        name: 'Lechmere - GLX',
```
digitalcora, line 2815: Should this just be `Lechmere`? It appears the other busway sign definitions use the plain station name, like `Nubian` or `Harvard`.

assets/js/mbta.ts
digitalcora: Ideally station signage would indicate the platforms as e.g. "Track 1" and "Track 2", then we would have official names for them. @jfabi do you know if that's the case?

assets/js/mbta.ts
PaulJKim: oops yes, I had made those but forgot to push them. Thanks for calling that out!

assets/js/mbta.ts
PaulJKim: Oh yeah, we should be able to do that since we deleted the old Lechmere busway!

assets/js/mbta.ts
jfabi: Yes, **Union Square** already has the identifiers "Track 1" and "Track 2" installed as follows (photo taken from Prospect Street):

DSC_0427

**At Lechmere**, the platforms are not yet signed up, but when they are installed, they will read "(D) Union Square / (E) College Ave" (on the same track as Track 1 at Union Square), and "Copley & West" (on the same track as Track 2 at Union Square).

PaulJKim: Thank you for the insight here @jfabi and @digitalcora , I was stuck on what to name these for a bit since they're both "westbound" tracks. Just pushed a change to reflect the correct track identifiers here and if it looks good to you @digitalcora , I will go ahead and make the corresponding change on the realtime signs end!

assets/js/mbta.ts
```diff
@@ -1987,16 +1987,65 @@ const stationConfig: {
...
+            },
+          },
+        },
+      },
       {
         id: 'GLEC',
         name: 'Lechmere',
         zones: {
+          e: {
+            label: 'Union Sq/College Ave',
```
digitalcora, line 2027: Does this fit into the UI? We might want something shorter. (I assume it's fine if it's `Union Sq` for now, until the rest of GLX opens)

assets/js/mbta.ts
jfabi: Just for the record (because I was originally going off of blueprints without thinking), it should be "(D) Union Square / (E) Medford/Tufts", since College Ave station has since been renamed Medford/Tufts. (Aside: I wonder how the slashes will work there.)

This shouldn't affect this pull request as you've dropped "College Ave" entirely from the label.

digitalcora (approved):
Seems ok to me to start testing in dev! Of course, if you can wait until tomorrow morning, @losvedir could also give it a final once-over.

PaulJKim:
Thanks for the review @digitalcora and sounds like a good plan to me, will wait for final approval from @losvedir!

assets/js/mbta.ts
```diff
@@ -3135,15 +3169,19 @@ const arincToRealtimeIdMap: { [key: string]: string } = {
...
+  'GUNS-m': 'union_sq_mezzanine',
   'SDUD-n': 'bus.Nubian_Platform_F',
   'SDUD-s': 'bus.Nubian_Platform_C',
   'SDUD-e': 'bus.Nubian_Platform_D',
   'SDUD-w': 'bus.Nubian_Platform_A',
   'SDUD-c': 'bus.Nubian_Platform_E_east',
   'SDUD-m': 'bus.Nubian_Platform_E_west',
-  'GLEC-n': 'bus.Lechmere_bus_north',
-  'GLEC-m': 'bus.Lechmere_bus_south',
+  'SLEC-m': 'bus.Lechmere_bus_mezzanine',
```
losvedir, line 3184: Did you talk to Dan Salomon about these changes? These IDs come from TransitWay Engine which he maintains.

losvedir (approved):
🚀 🎉

assets/js/mbta.ts
PaulJKim: Not yet - will reach out and make sure we're on the same page about the ids!

assets/js/mbta.ts
```diff
@@ -1987,16 +1987,65 @@ const stationConfig: {
...
+              auto: false,
               custom: true,
-              headway: true,
+              headway: false,
+              off: true,
+            },
+          },
+          m: {
+            modes: {
+              auto: false,
```
dks-mbta, line 2046: Should we have `label: 'Busway',` ?

assets/js/mbta.ts
```diff
@@ -1987,16 +1987,65 @@ const stationConfig: {
...
             modes: {
-              auto: true,
+              auto: false,
               custom: true,
-              headway: true,
+              headway: false,
+              off: true,
+            },
+          },
+          m: {
```
dks-mbta, line 2044: For consistency, this should be under SLEC rather than GLEC.

assets/js/mbta.ts
```diff
@@ -2762,27 +2811,12 @@ const stationConfig: {
...
-            label: 'North',
-            modes: {
-              auto: true,
-              custom: true,
-              headway: false,
-              off: true,
-            },
-          },
           m: {
-            label: 'South',
```
dks-mbta, line 2783: Should this include `label: 'Busway',` ?

assets/js/mbta.ts
dks-mbta: Ultimately this has to agree with ARINC's zone assignments. If the plan is to have the busway be SLEC m, this is fine with me, and I'll configure the Transitway Engine to agree with this. Will there also be sign(s) in the GLEC m zone for rail predictions?

Are these configured as `auto: false,` and `headway: false,` to lock off the signs because revenue service hasn't started?

PaulJKim: That's correct - SLEC-m is for buses and there is also GLEC-m for rail.

And yep, the plan currently is to follow up and unlock those signs for `headway` and `auto` mode once revenue service begins

assets/js/mbta.ts
PaulJKim: Hmm yeah probably a good idea. Although the UI does have an overarching **Busway.** label at the top of this page, 'Mezzanine' is kind of ambiguous in this context so Busway is probably clearer than that

PR #727, opened 2022-03-03
Add Braintree busway sign

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Reconfigure Braintree Busway Sign](https://app.asana.com/0/1180642489429371/1180642489429379/f)

This PR adds the config for the signs at the Braintree Busway. The sign has been offline for some time, but we are having ARINC bring it back on. Because the Center and Mezzanine zones are already being used for Subway predictions, ARINC is mapping the signs to the North zone which is why they are configured as `SBRA-n` in Signs UI.

### DIFF
--- a/assets/js/mbta.ts
+++ b/assets/js/mbta.ts
@@ -2922,6 +2922,26 @@ const stationConfig: {
           },
         },
       },
+      {
+        id: 'SBRA',
+        name: 'Braintree',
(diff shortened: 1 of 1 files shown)


dks-mbta (approved):
Looks good, pending zone confirmation from ARINC and passing formatting.

PR #739, opened 2022-03-22
Set Lechmere busway sign to have auto mode enabled

PaulJKim:
#### Summary of changes
This PR simply sets this sign to have auto mode enabled.

Context: These signs actually don't exist in real life. But there are bus predictions being provided technically, so for now we are at least enabling this in dev for the time being to monitor predictions through Signs UI.


cmaddox5 (approved):
👍

PR #750, opened 2022-04-11
Add ecs deploy workflows

PaulJKim:
#### Summary of changes
This PR adds GitHub Actions workflows for deploying Signs UI to ECS

### DIFF
--- a/.github/workflows/dev-ecs.yml
+++ b/.github/workflows/dev-ecs.yml
@@ -24,10 +24,14 @@ jobs:
         uses: mbta/actions/commit-metadata@v1
       - uses: mbta/actions/build-push-ecr@v1
         id: build-push
+        env:
+          ERL_COOKIE: ${{ secrets.ERL_COOKIE }}
+          SECRET_KEY_BASE: ${{ secrets.SECRET_KEY_BASE }}
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


PaulJKim:
@losvedir mainly adding you here because I noticed there are some secrets in the Dockerfile and EB workflow called `secrets.ERL_COOKIE` and `secrets.SECRET_KEY_BASE`. I'm unclear on what these are for and would we still need them in the ECS workflow?

PaulJKim:
@losvedir mainly adding you here because I noticed there are some secrets in the Dockerfile and EB workflow called `secrets.ERL_COOKIE` and `secrets.SECRET_KEY_BASE`. I'm unclear on what these are for and would we still need them in the ECS workflow?

PaulJKim:
Adding @cmaddox5 since you said you didn't have too much on your plate at the moment 😄

cmaddox5 (approved):
This looks good to me! I don't have a ton of experience with this, but comparing this to ECS deploy workflows that I know work currently, this should do the trick.

PaulJKim:
Thank you for clarifying that! I think it makes sense to move those configs to Terraform/the elixir releases in a separate PR since it seems like making a change to the releases would affect any potential deploys we would have to make in the meantime to EB.

For now, I'll just add them to the workflows like they were in the EB deploy workflow. Lmk if you think otherwise though!

PR #762, opened 2022-04-25
Migrate to native releases

PaulJKim:
Migrate Signs UI to use mix release instead of Distillery.

Moving to native Elixir releases removes the need for the ERL_COOKIE secret and lets us package the application without the use of Distillery.


Dockerfile
```diff
@@ -53,7 +49,11 @@ ENV MIX_ENV=prod TERM=xterm LANG="C.UTF-8" PORT=4000
...
 # add frontend assets with manifests from app build container
 COPY --from=app-builder /root/priv/static ./priv/static
-COPY --from=app-builder /root/_build/prod/rel/signs_ui/releases/current/signs_ui.tar.gz .
+COPY --from=app-builder /root/_build/prod/rel/signs_ui .
 RUN mkdir gtfs
-RUN tar -xzf signs_ui.tar.gz
-CMD ["bin/signs_ui", "foreground"]
+
+# Healthcheck
+HEALTHCHECK CMD ["bin/signs_ui", "rpc", "1 + 1"]
```
jzimbel-mbta, line 56: [TIL](https://hexdocs.pm/mix/1.13/Mix.Tasks.Release.html#module-one-off-commands-eval-and-rpc)

So does `HEALTHCHECK` know to expect this to print "2"?

.github/workflows/dev-ecs.yml
```diff
@@ -24,9 +24,6 @@ jobs:
         uses: mbta/actions/commit-metadata@v1
       - uses: mbta/actions/build-push-ecr@v1
         id: build-push
-        env:
-          ERL_COOKIE: ${{ secrets.ERL_COOKIE }}
```
jzimbel-mbta, line 28: Is this simply not needed anymore? Looks like the code that sets it here, as well as the code that uses it in the Dockerfile, was removed.

.github/workflows/dev-ecs.yml
```diff
@@ -24,9 +24,6 @@ jobs:
...
-          SECRET_KEY_BASE: ${{ secrets.SECRET_KEY_BASE }}
```
jzimbel-mbta, line 29: Should the `docker-additional-args` value a few lines down from here be updated to not include `--build-arg ERL_COOKIE --build-arg SECRET_KEY_BASE` since these have been removed?

Dockerfile
```diff
@@ -2,10 +2,6 @@
 FROM hexpm/elixir:1.11.4-erlang-23.3.1-debian-buster-20210326 as elixir-builder
 ENV LANG="C.UTF-8" MIX_ENV="prod"
 
-ARG ERL_COOKIE
-ENV ERL_COOKIE=${ERL_COOKIE}
-RUN if test -z $ERL_COOKIE; then (>&2 echo "No ERL_COOKIE"); exit 1; fi
-
 ARG SECRET_KEY_BASE
 ENV SECRET_KEY_BASE=${SECRET_KEY_BASE}
```
jzimbel-mbta, line 6: If this is no longer set in the GH workflow yaml, where does it come from?

jzimbel-mbta (commented):
I'm assuming this is good to go if it appears to work for you—I mostly just have questions since I'm still relatively inexperienced with the finer details of Docker/build scripts.

.github/workflows/dev-ecs.yml
PaulJKim: Yep, apparently using Elixir releases eliminates the need to set this explicitly and it's generated randomly with `mix release`. It's used by Distillery at build time and I don't have too many details, but [this PR comment](https://github.com/mbta/signs_ui/pull/750#issuecomment-1096923145) as well as [this one](https://github.com/mbta/devops/pull/329#issuecomment-1100128716) mentions that its not needed with native releases.

I _think_ it's the value generated at `_build/prod/rel/signs_ui/releases/COOKIE` which is used in the shell script that's called to start the app

.github/workflows/dev-ecs.yml
PaulJKim: oh yes, good call. I'll remove those!

Dockerfile
PaulJKim: I believe it's generated and packaged with the release itself so there isn't a need to read it in from an env variable

Dockerfile
jzimbel-mbta: Ah, Docker just checks the exit status of the command, not anything that it prints out. So this could print anything, and would be considered healthy as long as it has an exit status of 0.
https://docs.docker.com/engine/reference/builder/#healthcheck

PaulJKim: Hmm.. honestly am not certain. I added this while I was sort of "stabbing in the dark" with things that would help me understand why I couldn't run the image locally or deploy. While it doesn't seem to hurt to have it, I think I'm just going to remove it since it wasn't there prior and I'm also not too familiar with messing with Dockerfiles and also shell scripting.

jzimbel-mbta (approved):
Looks good to me, but it definitely would be a good idea to get more than just my eyes on it!

PR #763, opened 2022-04-26
Add missing bracket to ECS GitHub workflow

PaulJKim:
Was missing a bracket


cmaddox5 (approved):
We've all been there 😆

PR #807, opened 2022-08-02
Move screenplay URLs to prod.exs

PaulJKim:
The initial configuration didn't work when Signs UI was deployed to dev in ECS since the `MIX_ENV` used is still prod and so `dev.exs` isn't actually evaluated in dev ECS.

Moved the config to `prod.exs` instead and used the system env variables to branch depending on which ECS environment we're deployed to.

Also including localhost as an accepted frame ancestor so that local instances are able to embed Signs UI Dev renderings. Otherwise while doing local development, one would have to start up signs UI locally alongside Screenplay in order for the renderings to be viewable.

### DIFF
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -1,5 +1,4 @@
 import Config
-use Logger
 
 # For production, we often load configuration from external
 # sources, such as your system environment. For this reason,
--- a/config/prod.exs
+++ b/config/prod.exs
@@ -1,4 +1,5 @@
 import Config
+use Logger
 
 # For production, we often load configuration from external
 # sources, such as your system environment. For this reason,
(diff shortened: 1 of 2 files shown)


config/prod.exs
```diff
@@ -36,6 +35,16 @@ config :ueberauth, Ueberauth.Strategy.Cognito,
   user_pool_id: {System, :get_env, ["COGNITO_USER_POOL_ID"]},
   aws_region: {System, :get_env, ["COGNITO_AWS_REGION"]}
 
+if System.get_env("ENVIRONMENT_NAME") == "dev" do
+  config :signs_ui, SignsUiWeb.Endpoint,
+    screenplay_base_url: "localhost:4000/ https://screenplay-dev.mbtace.com/"
```
cmaddox5, line 40: Is the `localhost:4000/` here intentional or a copy/paste error?

PaulJKim: @cmaddox5 it's intentional but lmk what your thoughts are. Basically, I added localhost so that when you run screenplay locally, Signs UI Dev will let it embed the PAESS renderings. Otherwise, one would need to have a local instance of Signs UI running alongside Screenplay in order for things to fully work

cmaddox5: Ahh I see. That makes sense. Especially for the read-only phase. When we start implementing overrides, we may want to ensure that we are not locally testing overrides on alerts outside of local development. But in the meantime, this will be helpful.

cmaddox5 (approved):

PR #809, opened 2022-08-03
Add frame ancestors to content-security-policy at controller

PaulJKim:
Third times the charm. 

Basically, the prior attempts at a solution didn't work because we were depending on runtime values at compile time. This should work since we check for the `ENVIRONMENT_NAME` in `runtime.exs` and we append to the response header just before the controller returns a response.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -6,10 +6,10 @@ if config_env() == :prod do
 end
 
 
-if System.get_env("ENVIRONMENT_NAME") == "prod" do
-  config :signs_ui, SignsUiWeb.Endpoint,
-    screenplay_base_url: "https://screenplay.mbta.com/"
-else
-  config :signs_ui, SignsUiWeb.Endpoint,
-    screenplay_base_url: "localhost:4000/ https://screenplay-dev.mbtace.com/"
-end
+screenplay_base_url =
+  if System.get_env("ENVIRONMENT_NAME") == "prod",
+    do: "https://screenplay.mbta.com/",
+    else: "localhost:4000/ https://screenplay-dev.mbtace.com/"
+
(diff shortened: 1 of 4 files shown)


config/runtime.exs
```diff
@@ -4,3 +4,12 @@ if config_env() == :prod do
...
 end
+
+
+if System.get_env("ENVIRONMENT_NAME") == "prod" do
+  config :signs_ui, SignsUiWeb.Endpoint,
+    screenplay_base_url: "https://screenplay.mbta.com/"
+else
+  config :signs_ui, SignsUiWeb.Endpoint,
+    screenplay_base_url: "localhost:4000/ https://screenplay-dev.mbtace.com/"
+end
```
jzimbel-mbta, line 15: Is the fallback URL right? "localhost:4000/ https://screenplay-dev.mbtace.com/"

Also, if you want, you could get the value separately just to avoid some duplicated code:
```suggestion
screenplay_base_url =
  if System.get_env("ENVIRONMENT_NAME") == "prod",
    do: "https://screenplay.mbta.com/",
    else: "localhost:4000/ https://screenplay-dev.mbtace.com/"

config :signs_ui, SignsUiWeb.Endpoint,
    screenplay_base_url: screenplay_base_url
```

jzimbel-mbta (commented):
Looks good overall, just one comment.

config/runtime.exs
PaulJKim: Yep the fallback URL is right. You can add one or more `frame-ancestors` with just URLs separated by whitespace. The reason I have both localhost and Signs UI Dev is so that both your local instance of Screenplay along with dev will be allowed to embed the Signs UI URLs

jzimbel-mbta (approved):

PR #816, opened 2022-08-12
Allow parentheses

PaulJKim:
This change allows Signs UI to parse parentheses and allow them in custom text messages as well. The need for this came out of the JFK/UMass improvements that will use parentheses for communicating what platform the next northbound train will arrive at.

Dan Salomon reported that parentheses can work on countdown clocks in [this thread](https://mbta.slack.com/archives/G01HD71N1GD/p1652378170981619?thread_ts=1652376450.649179&cid=G01HD71N1GD) and I also tested this in the real world at the Silver Line Courthouse station mezzanine as pictured below
Screenshot_20220812-112330_2

### DIFF
--- a/assets/js/SignTextInput.tsx
+++ b/assets/js/SignTextInput.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 function isValidText(text: string) {
-  return !/[^a-zA-Z0-9,/!@' +]/.test(text);
+  return !/[^a-zA-Z0-9,/!@()' +]/.test(text);
 }
 
 interface SignTextInputProps {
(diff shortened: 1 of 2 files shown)


assets/js/SignTextInput.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 function isValidText(text: string) {
-  return !/[^a-zA-Z0-9,/!@' +]/.test(text);
+  return !/[^a-zA-Z0-9,/!@()' +]/.test(text);
```
jzimbel-mbta, line 4: https://regex101.com/ suggests escaping the `/` in this pattern (so, `\/`), but when I pasted it into my browser console it worked, so... seems fine :shrug:

I'm also a little confused by the double-negative (from `!` and the `[^]` inverted character class), but again if it works it works.

jzimbel-mbta (approved):
Good to go!

assets/js/SignTextInput.tsx
PaulJKim: Oh I didn't even notice that. It is kinda weird.. but yeah I guess if it's already in the codebase and working without issue then ¯\_(ツ)_/¯

PR #831, opened 2022-09-08
Add recurring task to keep S3 config clean of unused keys

PaulJKim:
#### Summary of changes
**Asana Ticket:** [New Feature - improve updating of config in S3](https://app.asana.com/0/1201753694073608/1201722525356149/f)

This PR moves the ARINC sign Id to Realtime Signs Id mapping from the frontend in the `mbta.ts` file to the backend in a JSON file. The reasoning behind this was so that the backend could know about all of the signs that _should_ exist on Signs UI and reference that in order to keep the S3 config clean by periodically (once a minute) running a task to check all of the keys in the S3 config and remove any that are not in the local JSON file.

I'm open to thoughts on the approach. The mapping of ids rarely changes so maybe the cleaning task can be more ad-hoc. For example it could be triggered through a web endpoint whenever needed. I went with the once-per-minute loop though just to keep things more hands-off.


PaulJKim:
Didn't expect to run into the failing Sobelow check. Looking into that now

lib/signs_ui/config/utilities.ex
```diff
@@ -0,0 +1,28 @@
+defmodule SignsUi.Config.Utilities do
+  @moduledoc """
+  Utility functions for interacting with the arinc_to_realtime.json file
+  """
+
+  def clean_configs(sign_configs) do
+    registered_ids = get_realtime_ids()
+
+    :maps.filter(
```
jzimbel-mbta, line 9: Depending on which version of elixir signs_ui is running, you might be able to use `Map.filter/2` for this

lib/signs_ui/config/utilities.ex
```diff
@@ -0,0 +1,28 @@
...
+      fn key, _ ->
+        key in registered_ids
+      end,
+      sign_configs
+    )
+  end
+
+  def get_realtime_ids do
+    get_arinc_to_realtime_mapping() |> Enum.map(fn {_k, v} -> v end)
```
jzimbel-mbta, line 18: If `get_arinc_to_realtime_mapping()` returns a map, you could use `Map.values/1` instead of the current `Enum.map`

assets/js/helpers.ts
```diff
@@ -104,4 +104,25 @@ function defaultZoneLabel(z: Zone): string {
   }
 }
 
-export { choosePage, alertsForLine, formatTime, defaultZoneLabel };
+function arincToRealtimeId(
+  stationZone: string,
+  line: string,
+  arincToRealtimeIdMap: { [key: string]: string },
```
jzimbel-mbta, line 110: If `arincToRealtimeIdMap` is passed to the client as essentially a global constant (for the lifetime of the page) on the `window` object, does this function need to receive it as an argument? Does it need to be passed through the component tree as props?

If we don't expect it to ever change for the client's lifetime, I think it would be ok for the `return` line of this function to simply `return window.arincToRealtimeIdMap[stationZone];`.

jzimbel-mbta (approved):
Looks good to me, just some questions/suggestions on the approach.

Not sure why Sobelow is complaining about your `File.read!` call, but I'm approving assuming you address it and get CI checks passing again.

lib/signs_ui/config/utilities.ex
PaulJKim: Signs UI is currently running on Elixir 1.11 so Map.Filter isn't available. We should probably update the version though soon

assets/js/helpers.ts
PaulJKim: Ah yep you're totally right and I agree. This simplifies things a lot!

PR #835, opened 2022-09-12
Remove outdated logic for RL predictions for SL transfers

PaulJKim:
#### Summary of changes
This PR is a minor tweak of #831. The `red_south_station_mezzanine` id is used for the actual South Station Red Line Mezzanine sign, which always shows Red Line predictions. The `south_station_mezzanine` id was _intended_ for a sign at the Silver Line platform that shows Red Line predictions for riders that are transferring from Silver to Red. It turns out that there is already an Id called `south_station_silver_line_arrival_platform` for this same use-case so the id `south_station_mezzanine` is actually not being used and is defunct. Therefore, we can safely remove the dedicated logic in the helper function.

I also considered removing the helper function entirely, but I think keeping the function there is a little more readable.

### DIFF
--- a/assets/js/Line.tsx
+++ b/assets/js/Line.tsx
@@ -57,7 +57,7 @@ function setAllStationsMode(
 
   stations.forEach((station: StationConfig) => {
     ['n', 's', 'e', 'w', 'm', 'c'].forEach((zone) => {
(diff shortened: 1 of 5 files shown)


jzimbel-mbta (approved):
I agree that even though it's very simple now, it's best to keep the helper function around, just to minimize the number of places where we mess with global `window.*` variables.

PR #929, opened 2023-02-07
remove improper hook call

panentheos:
#### Summary of changes
**Asana Ticket:** [Chore: Update error-prone call to hook in Station component](https://app.asana.com/0/1201753694073608/1201950647334669/f)

This fixes a situation where a hook was being called inside a conditional, which is not valid and can cause bugs. Since downstream components aren't memoized, this `useCallback` wasn't providing any optimizations, so removing it was the simplest option.

### DIFF
--- a/assets/js/Station.tsx
+++ b/assets/js/Station.tsx
@@ -67,12 +67,9 @@ function makeSign(
     const signGroup = signGroupKey ? signGroups[signGroupKey] : undefined;
     const ungroupMe = signGroupKey ? () => ungroupSign(realtimeId) : undefined;
 
-    const setConfig = React.useCallback(
-      (conf: SignConfig) => {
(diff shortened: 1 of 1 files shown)


jzimbel-mbta (approved):
I've actually never used the `useCallback` hook! (or `useMemo` for that matter)

Even if this hook were used at the top level and not incorrectly inside a conditional, would it be a good use of the hook? It doesn't seem like the callback does anything expensive, unless `setConfigs` would cause a very large re-render or something.

panentheos:
The main use of `useCallback` is in conjunction with `memo` to optimize re-renders of downstream components. Since we're not using `memo` here, this `useCallback` isn't doing much. Generally, that kind of optimization isn't needed most of the time, and it can be somewhat tricky to set up, so it's best to skip it unless there's an obvious performance issue.

PR #930, opened 2023-02-08
fix crash in AlertPicker

panentheos:
#### Summary of changes
**Asana Ticket:** [Crash when alert picker dropdown is open while an alert is closed](https://app.asana.com/0/1201753694073608/1203852555384568/f)

This fixes a crash when the last alert for a line is removed while the `AlertPicker` is open. The `key` for `AlertPickerPopup` changes every time the list of alerts changes, causing the popup to re-mount and reinitialize the selected alert id to the first in the list. When the new list is empty, the selected id becomes undefined, which later causes the crash. This changes the guard clause to ensure we actually have an alert before accessing it.

### DIFF
--- a/assets/js/AlertPicker.tsx
+++ b/assets/js/AlertPicker.tsx
@@ -17,7 +17,7 @@ function AlertPickerPopup({
   const [hoveredAlertId, setHoveredAlertId] = React.useState<string>(
     alertIds[0],
   );
-  const hoveredAlert = alerts[hoveredAlertId];
+  const hoveredAlert = hoveredAlertId && alerts[hoveredAlertId];
 
   const ref = React.useRef<HTMLDivElement | null>(null);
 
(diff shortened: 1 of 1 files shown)


jzimbel-mbta (approved):
This works, but do you need to have the `hoveredAlertId &&` in the assignment on line 20? If the value of that variable is `null` or `undefined`, I would assume that `alerts[hoveredAlertId]` will give you `undefined` anyway, which would not pass the guard.

Another fix that would be a one-liner, I think, would be to change the equality check on line 58/59 from `===` to `==`. One of the few good use cases for `==` is when checking whether a value is either `null` or `undefined`.

panentheos:
Correction to the above: Actually, regarding the first suggestion, note that `obj[undefined]` effectively becomes `obj["undefined"]`, which is certainly not what we want to do (even if it probably would't find anything due to the current naming scheme for alerts).

PR #949, opened 2023-02-14
improve timings of paging messages

panentheos:
#### Summary of changes

This makes two improvements to the display of paging messages:

Because timeout callbacks are always approximate, the interval between `currentTime` updates is usually slightly above or below 1 second. We effectively truncate the time in `choosePage`, causing random 1 sec variations in paging duration. By rounding first, we can eliminate that variation. Also fixes an off-by-one error.

ARINC signs have the quirk of starting paging on page 2 of a message. We compensate for this in realtime-signs by shifting the pages so the desired one is shown first. This change emulates the sign behavior by shifting the other way, so the net result is the correct ordering.

### DIFF
--- a/assets/js/SignPanel.tsx
+++ b/assets/js/SignPanel.tsx
@@ -87,7 +87,11 @@ function lineDisplayText(
     lineContent !== undefined &&
     isNotExpired(lineContent.expiration, currentTime)
   ) {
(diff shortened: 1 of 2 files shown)


PaulJKim (approved):
LGTM! Thank you for looking into this!

PR #989, opened 2023-03-01
Replace Enzyme part 1

PaulJKim:
#### Summary of changes
**Asana Ticket:** [[Chore] Remove enzyme library](https://app.asana.com/0/1201753694073608/1203971638649519/f)

This PR removes the enzyme library from Signs UI. This library was abandoned several years ago and react testing library is favored as it's replacement.

To avoid PR bloat, I'll put this first half of the tests up first. Then, I'll finish up the last half and fully remove the library in a subsequent PR.


assets/test/Line.test.ts
```diff
@@ -604,19 +607,43 @@ test('can toggle chelsea bridge announcements on and off on Silver Line page', (
...
   );
 
-  const chelseaInput = wrapper.find('input[name="chelsea_bridge"]');
-  chelseaInput.simulate('change', { target: { checked: true } });
+  fireEvent.click(screen.getByTestId('chelsea_bridge_toggle'));
   expect(setChelseaBridgeAnnouncements.mock.calls.length).toEqual(1);
   expect(setChelseaBridgeAnnouncements).toHaveBeenCalledWith('auto');
+});
+
+test('can toggle chelsea bridge announcements off on Silver Line page', () => {
```
panentheos, line 615: Does this need to be a separate test? It's not always possible, but chaining together multiple interactions can increase the effectiveness of UI tests by potentially catching more complex bugs (with less code, too).

panentheos (commented):
This is great, thanks for picking this up! See a couple comments.

assets/test/ConfiguredHeadwaysForm.test.ts
```diff
@@ -29,21 +28,6 @@ beforeEach(() => {
...
-      {
-        branches,
-        readOnly,
-        configuredHeadways,
-        setConfiguredHeadways,
-        timePeriods,
-      },
-      null,
-    ),
-  );
```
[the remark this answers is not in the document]
PaulJKim: Ah yeah that's true. I think that was just because this was the first file I migrated and was still figuring stuff out. Will adjust accordingly!

assets/js/SetExpiration.tsx
```diff
@@ -56,6 +56,7 @@ function SetExpiration({
       {(!readOnly && (
         <label>
           <input
+            data-testid="set_return_to_auto_checkbox"
```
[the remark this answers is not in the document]
PaulJKim: ohh gotcha, thank you for the tip!

assets/test/Line.test.ts
PaulJKim: Ah I think I might have split this one up because for some reason when using `fireEvent` or `userEvent` to simulate the toggle going off and on, the `setChelseaBridgeAnnouncements` mock was getting called with `auto` both times for some reason. Any thoughts on why that might be happening? I also know that this is more testing implementation details which I understand sort of goes against the principles of RTL. I feel like with this toggle, we may still want to be testing for those details, but if you have other thoughts, I'd be interested in hearing them!

PaulJKim: Actually scratch that, I think what I described above happens because of how the onChange handler is mocked. The handler is passed down from the top in `ViewerApp` and is supposed to broadcast to all of the phoenix channels first and then rerenders from there.

assets/test/Line.test.ts
```diff
@@ -52,7 +50,7 @@ test('Shows all signs for a line', () => {
   const readOnly = false;
   const configuredHeadways: ConfiguredHeadways = {};
 
-  const wrapper = mount(
+  render(
```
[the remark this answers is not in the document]
PaulJKim: Oof, yeah and those warnings aren't very specific or helpful at all. I'll figure out where we need to add `act` for now and we can add another chore task to investigate and potentially remove the library

assets/test/SetExpiration.test.ts
```diff
@@ -23,10 +23,10 @@ test('Can set the expiration time', () => {
     }),
   );
 
-  wrapper.find('input[type="checkbox"]').simulate('click');
-  wrapper.find('.set_expiration--datetime').simulate('change');
-  wrapper.find('.react-datepicker-wrapper input').simulate('focus');
-  wrapper.find('.react-datepicker__day--015').simulate('click');
+  await user.click(screen.getByText('Schedule return to "Auto"'));
+  await user.click(screen.getByText('Date and time'));
+  await user.click(screen.getByRole('option', { name: /^Choose.*15th.*/ }));
```
panentheos, line 28: This is great! This is the kind of expressiveness that we should be aiming for in our UI test code.

panentheos (approved):
Looks good! Thank you for taking this on, I know it's been a big lift.

assets/test/SetExpiration.test.ts
PaulJKim: Yeah took a bit of fiddling to get it to work the way I wanted it to, but love how it turned out compared to how it was before!

PR #992, opened 2023-03-02
Replace Enzyme part 2

PaulJKim:
#### Summary of changes
**Asana Ticket:** [[Chore] Remove enzyme library](https://app.asana.com/0/1201753694073608/1203971638649519/f)

This PR is part 2 to #989 and migrates the remaining tests still using Enzyme and then fully removes the library from the project. It also includes some minor changes to some `data-testid` attributes to make testing easier.

This branch was branched off of part 1 so those changes appear in the diff until part 1 is merged.


assets/test/SignPanel.test.ts
```diff
@@ -346,15 +371,82 @@ test.each([
...
+  );
+
+  expect(
+    screen.getByAltText('Line one custom text input').getAttribute('value'),
+  ).toBe('foo');
+  expect(
+    screen.getByAltText('Line two custom text input').getAttribute('value'),
+  ).toBe('bar');
+
+  await user.selectOptions(
```
PaulJKim, line 434: These two tests were in `ViewerApp.test.tsx` but were just testing a single sign panel so I thought it made more sense to move them into this file

assets/test/ViewerApp.test.tsx
```diff
@@ -177,5 +109,5 @@ test('Shows sign out link', () => {
     ),
   );
 
-  expect(wrapper.find('#sign-out-link').props().href).toBe('/path');
+  expect(screen.getByRole('link')).toHaveAttribute('href', '/path');
```
panentheos, line 112: Since this selector is so general, this test would fail if we ever added another link on the page. We should add the link text to the query here (which will also make the test clearer to read).

panentheos (approved):
See one small comment, then looks good. Glad to see those dependencies removed!

PR #1023, opened 2023-03-14
Remove Formik

PaulJKim:
#### Summary of changes
**Asana Ticket:** [[Chore] Remove formik library](https://app.asana.com/0/1201753694073608/1203971638649517/f)

This PR removes Formik and replaces it with React Hook Forms. React Hook Forms is compatible with React 18 and actively maintained, unlike Formik. It is also a good deal simpler than Formik while supporting a lot of the same or similar features.

Couple of notes:
- There appear to be some breaking changes for react hook forms with `yup` v1. This required me to downgrade to 0.32.11 for the time being.
- React hook forms encourages props spreading which is discouraged by eslint

### DIFF
--- a/assets/js/ConfiguredHeadwaysForm.tsx
+++ b/assets/js/ConfiguredHeadwaysForm.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { Formik, Form, Field, FieldArray } from 'formik';
+import { useForm, SubmitHandler } from 'react-hook-form';
+import { yupResolver } from '@hookform/resolvers/yup';
 import { object, string, array, number, ref } from 'yup';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faEdit, faTimes } from '@fortawesome/free-solid-svg-icons';
@@ -7,6 +8,18 @@ import { timePeriodConfig } from './mbta';
 import { ConfiguredHeadways } from './types';
 
 /* eslint-disable camelcase */
+
+type Inputs = {
+  branches: {
+    id: string;
+    [timePeriodId: string]:
+      | {
+          range_high: number;
+          range_low: number;
+        }
+      | string;
+  }[];
+};
 
 interface ConfiguredHeadwaysFormProps {
   branches: {
@@ -87,7 +100,7 @@ function ConfiguredHeadwaysForm({
     setInEditMode(!isEnabled);
   }, [branches, configuredHeadways, isEnabled]);
 
-  const handleSubmit = React.useCallback(
+  const onSumbit: SubmitHandler<Inputs> = React.useCallback(
     (values: {
       branches: {
         id: string;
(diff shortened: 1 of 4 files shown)


assets/js/ConfiguredHeadwaysForm.tsx
```diff
@@ -113,128 +138,98 @@ function ConfiguredHeadwaysForm({
...
-            | string;
-        }[];
-      }>
-        validationSchema={formSchema}
-        enableReinitialize
-        initialValues={initialFormValues}
-        onSubmit={handleSubmit}
-        validateOnBlur
-        validateOnChange={false}
+      {errors.branches ? (
```
panentheos, line 147: I just tested this locally, and noticed some odd behavior around when the error message displays. Sometimes it doesn't show up after making an invalid edit to one of the two fields, and sometimes it persists even after the form transitions to valid. I didn't dig in too much, but I suspect it might be related to how we're using `errors.branches` here. Maybe this should be `!isValid` instead?

assets/package.json
```diff
@@ -57,13 +57,14 @@
...
     "react-dom": "^16.4.1",
+    "react-hook-form": "^7.43.5",
     "source-map-loader": "^4.0.1",
     "ts-jest": "^29.0.5",
     "ts-loader": "^9.4.2",
     "typescript": "^4.9.5",
     "webpack": "^5.75.0",
     "webpack-cli": "^5.0.1",
-    "yup": "^1.0.2"
+    "yup": "^0.32"
```
panentheos, line 67: Why the downgrade?

assets/test/ConfiguredHeadwaysForm.test.ts
```diff
@@ -284,27 +296,21 @@ describe('With Multi-sign Headway enabled', () => {
...
+        '10',
       );
-
       await user.type(
         container.querySelector(
-          'input[name="branches.[0].morning.range_low"]',
+          'input[name="branches.[0].morning.range_high"]',
         )!,
-        '10',
+        '8',
```
panentheos, line 307: Were these switched for clarity, or because the order matters?

assets/js/ConfiguredHeadwaysForm.tsx
PaulJKim: Ah yep I think you're right. Changed it to use `!isValid` and seems to work much more consistently. I think I opted to not use that flag at first because when running locally, due to the absence of values, the flag will show up immediately but that actually would be the behavior we want since all the values are required by the schema. All that said, this change would actually be somewhat of a bugfix considering that the current version of Signs UI doesn't show the error banner when running local.

assets/test/ConfiguredHeadwaysForm.test.ts
PaulJKim: Order did matter because of the quirk that you mentioned in your prior comment about the alert banner sometimes not showing up. But I just switched it back since the form behaves better with the change to use `isValid` as the condition for showing the error banner.

assets/package.json
PaulJKim: yup 1.0 was released fairly recently and had breaking changes that the `@hookform/resolvers` doesn't support yet

assets/js/ConfiguredHeadwaysForm.tsx
panentheos: I'm confused about that last bit. I've always seen the error banner when running locally.

In any case, I retested and it looks good now.

panentheos (approved):
Looks good, thanks for taking this on!

assets/js/ConfiguredHeadwaysForm.tsx
PaulJKim: Huh that's weird. For example, here's me running current master and then opening up the headways form:
https://user-images.githubusercontent.com/16074540/225113368-89f00eea-c84e-4660-aa24-3d2e815b8fc2.mov

panentheos: Ah, I misunderstood, I thought you meant you never saw the banner. That vid matches what I saw.

PR #1069, opened 2023-03-31
Add local playback script

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Playback mode in Signs UI](https://app.asana.com/0/1201753694073608/1204262777371382/f)

This PR adds a script to the Signs UI repo that can be used to read in a CSV of past realtime signs logs and replay them against a locally running copy of Signs UI.

### DIFF
--- a/scripts/local_playback/countdown_clock_playback.exs
+++ b/scripts/local_playback/countdown_clock_playback.exs
@@ -1,0 +1,100 @@
+# This script allows you to download a CSV file from Splunk containing logs of past sent messages
+# and play them back against a locally running instance of Signs UI.
+
+Mix.install([{:httpoison, "~> 1.0"}])
+
+if System.get_env("SIGN_UI_URL") == nil do
+  IO.puts("Please set the SIGN_UI_URL environment variable")
+  System.halt(1)
+end
+
+if System.get_env("SIGN_UI_API_KEY") == nil do
+  IO.puts("Please set the SIGN_UI_API_KEY environment variable")
+  System.halt(1)
+end
+
+defmodule CountdownClockPlayback do
+  use GenServer
+
+  @sign_ui_url System.get_env("SIGN_UI_URL")
+  @sign_ui_api_key System.get_env("SIGN_UI_API_KEY")
+
+  def start_link(csv_path) do
+    GenServer.start_link(__MODULE__, csv_path, name: __MODULE__)
+  end
+
+  @impl true
+  def init(csv_path) do
+    messages =
+      File.read!(csv_path)
+      |> String.split("\n")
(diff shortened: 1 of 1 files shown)


scripts/local_playback/countdown_clock_playback.exs
```diff
@@ -0,0 +1,100 @@
...
+  System.halt(1)
+end
+
+if System.get_env("SIGN_UI_API_KEY") == nil do
+  IO.puts("Please set the SIGN_UI_API_KEY environment variable")
+  System.halt(1)
+end
+
+defmodule CountdownClockPlayback do
+  use GenServer
```
panentheos, line 17: I haven't personally tried it yet, but I wonder if `Task` might be a good fit here. It would enable the code to be written as a single `Enum.each` with a `Process.sleep` for the delay, which might be slightly cleaner than the `Process.send_after` approach. The current approach is perfectly reasonable though, so consider this just a potentially interesting exploration.

scripts/local_playback/countdown_clock_playback.exs
```diff
@@ -0,0 +1,100 @@
...
+    messages =
+      File.read!(csv_path)
+      |> String.split("\n")
+      |> Stream.filter(fn row ->
+        String.contains?(row, "update_single_line:") or String.contains?(row, "update_sign:")
+      end)
+      |> Stream.map(fn row ->
+        [_, timestamp, _, _, update_type, message | _] = String.split(row, " ")
+        {:ok, timestamp, _} = DateTime.from_iso8601("#{timestamp}Z")
+        {timestamp, update_type, message}
```
panentheos, line 37: nit: `update_type` isn't needed below, so we could omit it

panentheos (approved):
Looks good! See a few optional comments.

scripts/local_playback/countdown_clock_playback.exs
PaulJKim: Hmm interesting, so would Task be used to encapsulate each http request? Usually `Task` would be used to take synchronous code and make it concurrent, but here we're essentially doing the opposite. Another reason I went with the GenServer approach was to have a little more flexibility in accessing the list since I'm looking ahead at the next message to calculate the delay. Using `Enum.each`, how would you suggest determining how much to sleep the process by?

panentheos: I was envisioning using a single `Task` for the whole simulation, similar to how you might use an `async` function in JS to perform a long-running sequence of calls. But yeah, getting the timing information makes it slightly more complicated, you'd need to use e.g. `Enum.chunk_by` to get each consecutive pair first.

PR #1078, opened 2023-04-06
formalize local config store

panentheos:
#### Summary of changes

This improves the local development experience by implementing proper local file storage for the sign config. Notes:
* The file lives at `priv/config.json`. If it does not exist on startup, it will be transparently created.
* Changes are persisted automatically, and the file is read on startup, so the system behaves more like production.
* The file (and all associated code) use the "new format". Removes support for the "old format" and some fallback code that seemingly hasn't been used in a long time (and looked incorrect in some cases).

This paves the way for wiring up a local realtime_signs to read the config directly from a local signs_ui.

Also disables the `PipeChainStart` credo rule, per external discussion


lib/signs_ui/config/state.ex
```diff
@@ -71,16 +72,22 @@ defmodule SignsUi.Config.State do
...
-        save_state(config)
-        schedule_clean(self(), 60_000)
-        {:ok, config}
+    schedule_clean(self(), 60_000)
+    config_store = Application.get_env(:signs_ui, :config_store)
+    response = config_store.read() |> Jason.decode!()
+
+    state = %{
+      signs:
+        Map.get(response, "signs", %{})
```
PaulJKim, line 81: Suggestion based on our slack convo: this is one spot where I might prefer to avoid the single pipe
```
response
|> Map.get("signs", %{})
|> Map.new(fn {sign_id, config} -> {sign_id, Sign.from_json(sign_id, config)} end)
```
for example just looks a little neater to me. But up to you if you like it as is

lib/signs_ui/config/state.ex
```diff
@@ -71,16 +72,22 @@ defmodule SignsUi.Config.State do
 
   @spec init(any()) :: {:ok, t()} | {:stop, any()}
```
PaulJKim, line 73: Minor: we can remove `{:stop, any()}` from the typespec since we'll never return that anymore

PaulJKim (approved):
Some minor suggestions but overall looks great! Thank you for doing this!

PR #1112, opened 2023-05-02
default to auto when no config

panentheos:
#### Summary of changes

This changes the default mode from 'off' to 'auto', if there is no explicit configuration for a sign, in order to match the behavior of realtime_signs. Practically, this only affects local development, since the deployed environments already have fully-populated configuration files, but it's still good for consistency.

### DIFF
--- a/assets/js/Line.tsx
+++ b/assets/js/Line.tsx
@@ -15,7 +15,7 @@ import {
   SignContent,
   StationConfig,
 } from './types';
-import { arincToRealtimeId } from './helpers';
+import { arincToRealtimeId, getSignConfig } from './helpers';
 
 /* eslint-disable camelcase */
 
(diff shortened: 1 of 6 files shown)


PaulJKim (commented):
The only thing I might call out here is that this means that when we add a new sign to Signs UI (rarely happens currently), it will show as in `auto` mode until we "instantiate" the config by changing the mode once so that it writes to S3.

Of course, it wouldn't _actually_ be in auto mode in this case, but it could look weird to someone who adds a new sign, sets auto mode to disabled, but then sees the sign as set to auto when they deploy if they didn't understand this defaulting behavior beforehand.

panentheos:
I'm not sure I understand. Realtime signs currently defaults to 'auto' when there's no config, so this change attempts to align with that. This also gives a slightly better experience if auto mode is locked out: It shows the option while it's default-selected, but once you switch away, it disappears so you can't go back. (Previously it erroneously showed 'Custom' because that happened to be the first one in the list)

PaulJKim:
Yeah that makes sense when relating this to how Realtime Signs defaults. But I don't think we'd ever want to add configs to Realtime Signs before first adding the signs to Signs UI. The typical workflow for adding a new sign is add the signs to Signs UI with all modes disabled except for `off`, then add the Realtime Signs configs once GTFS and RTR changes are deployed and enable auto and headway modes on Signs UI.

Hopefully what I'm trying to say makes more sense if I put it this way: With the logic in this PR, adding new signs to Signs UI and having them show as `auto` initially means that unless the person who deployed is an admin user of Signs UI prod (I think ideally we're not supposed to have admin access in prod for Signs UI as developers), they would have to ask PIOs to go into Signs UI and switch the signs to `off` for them and make the defaulted `auto` mode disappear.

I guess I'm coming at this more from the perspective of thinking about the deployed environments versus local development. But that's because I think would have some affect on those environments since there is technically a valid case for a sign to be on Signs UI but not have it's config populated.

panentheos:
We decided to change the default in realtime signs instead, closing this.

PR #1127, opened 2023-05-11
accept and process audio messages

panentheos:
#### Summary of changes
**Asana Ticket:** [Signs-UI shows a flag/toggle when audio has been sent to a zone](https://app.asana.com/0/1201753694073608/1204564881899174/f)

This allows the server to receive audio messages, parses and stores them, and then forwards them to the front-end. This is one piece of the overall feature, and will be followed by at least 2 more changes: sending the messages from realtime_signs, and updating the UI to actually display the information.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -41,7 +41,7 @@
         {Credo.Check.Refactor.MatchInCondition},
         {Credo.Check.Refactor.NegatedConditionsInUnless},
         {Credo.Check.Refactor.NegatedConditionsWithElse},
-        {Credo.Check.Refactor.Nesting},
+        {Credo.Check.Refactor.Nesting, false},
         {Credo.Check.Refactor.PipeChainStart, false},
         {Credo.Check.Refactor.UnlessWithElse},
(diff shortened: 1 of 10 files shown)


.credo.exs
```diff
@@ -41,7 +41,7 @@
         {Credo.Check.Refactor.MatchInCondition},
         {Credo.Check.Refactor.NegatedConditionsInUnless},
         {Credo.Check.Refactor.NegatedConditionsWithElse},
-        {Credo.Check.Refactor.Nesting},
+        {Credo.Check.Refactor.Nesting, false},
```
panentheos, line 44: FYI, I'm advocating for disabling this rule. While I agree with the principle of avoiding excessive nesting _when_ it's detrimental to code clarity, I think it should be a judgement call and not an arbitrary depth number.

PaulJKim: Yeah agreed that it's contrived to make it a hard and fast rule

PaulJKim (approved):
LGTM! Looking forward to how this turns out!

PR #1212, opened 2023-07-20
Allow colon

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Implement early AM prediction changes](https://app.asana.com/0/1201753694073608/1204795336662073/f)

We tested sending a message containing a colon directly via HTTP to the headend server and the message was posted to a sign. This PR allows Signs UI to render colons and also allows them to be used in custom messages as well.

image (4)


panentheos (approved):
Looks good.

PR #1220, opened 2023-07-25
disable/remove ungrouping buttons in read-only mode

panentheos:
#### Summary of changes
**Asana Ticket:** [Fix "Ungroup" button present on view/read-only mode](https://app.asana.com/0/1201753694073608/1205077966832802/f)

This removes access to the "Ungroup" and "Return to auto" buttons when in read-only mode. Note that I followed the rough pattern of removing the button entirely on the main page, and just disabling it within the dropdown form.


PaulJKim (approved):
LGTM!

PR #1264, opened 2023-09-07
Dev green workflow

PaulJKim:
#### Summary of changes
Proactively adding a new GH workflow for deploying to dev-green. Currently working with infra to get the environment stood up.

### DIFF
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -13,7 +13,7 @@ jobs:
   deploy:
     name: Deploy to Dev-green
     runs-on: ubuntu-20.04
-    environment: dev-green
+    environment: dev
     env:
       ECS_CLUSTER: signs-ui
       ECS_SERVICE: signs-ui-dev-green
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -1,0 +1,42 @@
+name: Deploy to Dev-green
+
+on:
+  workflow_dispatch:
+  push:
+    branches: master
+
+concurrency:
+  group: dev-green
+  cancel-in-progress: true
+
+jobs:
+  deploy:
+    name: Deploy to Dev-green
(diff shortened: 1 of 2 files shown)


.github/workflows/dev-green.yml
```diff
@@ -0,0 +1,42 @@
+name: Deploy to Dev-green
+
+on:
+  workflow_dispatch:
+  push:
+    branches: master
```
panentheos, line 6: This will automatically deploy on merges to master, right? Is that what we want for this env?

.github/workflows/dev-green.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  deploy:
+    name: Deploy to Dev-green
+    runs-on: ubuntu-20.04
+    environment: dev
+    env:
+      ECS_CLUSTER: signs-ui
+      ECS_SERVICE: signs-ui-dev-green
+
+    steps:
```
panentheos, line 21: Since these steps are (and presumably should be) identical to the other environments, is there an easy way to share the code between them?

panentheos (approved):
Looks good, see a few questions.

.github/workflows/dev-green.yml
PaulJKim: You're correct. Actually I don't think we want that and should only deploy to dev-green when manually triggered. Will remove the `push:` option

.github/workflows/dev-green.yml
PaulJKim: Oh okay, actually looks like there's a general "deploy to ECS" workflow for CTD that we could use to consolidate things https://github.com/mbta/workflows/blob/main/.github/workflows/deploy-ecs.yml

PaulJKim:
Added a generalized workflow for deploying to ECS to this PR. Will try that one out and if no issues, will follow up with a subsequent PR to remove the old workflows.

PR #1291, opened 2023-11-27
Poll alerts

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Rework Signs UI to direct poll for alert events](https://app.asana.com/0/1185117109217413/1205874282120116/f)

Replace the use of `GenStage` and Server Sent Events with polling the V3 API every 5 seconds for alerts data. With SSE, the events occasionally would stop coming through for a reason we were not able to pin down. Since Signs UI doesn't need to be fully "real-time", swapping out SSE for some simple polling should be more straightforward and reliable.


lib/signs_ui/alerts/state.ex
```diff
@@ -4,87 +4,147 @@ defmodule SignsUi.Alerts.State do
...
-    alert = Events.parse(data)
-    Map.put(state, alert.id, alert)
+  def parse(payload) when is_list(payload) do
+    Enum.map(payload, &parse/1)
   end
 
-  defp update_state(%Event{event: "add", data: data}, state) do
-    alert = Events.parse(data)
-    Map.put(state, alert.id, alert)
+  def parse(payload) when is_map(payload) do
```
panentheos, line 106: nit: Since these `parse` functions operate on different levels of the data structure, it might be slightly clearer to give them distinct names rather than implicitly differentiating by type. (And/or combine some of them)

lib/signs_ui/alerts/state.ex
```diff
@@ -4,87 +4,147 @@ defmodule SignsUi.Alerts.State do
...
-    GenStage.call(pid, :all)
+    GenServer.call(pid, :all)
   end
 
-  @impl GenStage
-  def init(opts) do
-    {:consumer, %{}, opts}
+  @impl GenServer
+  def init(_) do
+    Process.send(self(), :update, [])
```
panentheos, line 32: nit: could just use `send` since the third argument is empty.

lib/signs_ui/alerts/state.ex
```diff
@@ -4,87 +4,147 @@ defmodule SignsUi.Alerts.State do
...
+
+    case http_client.get(
+           url,
+           headers,
+           params: %{
+             "filter[datetime]" => "NOW",
+             "filter[route_type]" => "0,1"
+           }
+         ) do
+      {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
```
panentheos, line 90: Due to load balancing/caching, the API can sometimes return slightly older data on subsequent calls, which means that unless we track `Last-Modified`, we could see this state briefly "out of order" around transitions. That's probably fine for this case, since this data doesn't change very often, and it only populates the form UI, but I wanted to call it out.

lib/signs_ui/alerts/state.ex
```diff
@@ -4,87 +4,147 @@ defmodule SignsUi.Alerts.State do
...
+        {:ok, response} ->
+          new_state = Enum.into(parse(response), %{}, &{&1.id, &1})
+
+          SignsUiWeb.Endpoint.broadcast!(
+            "alerts:all",
+            "new_alert_state",
+            Display.format_state(new_state)
+          )
+
+          Logger.info(["alert_state_updated ", inspect(new_state)])
```
panentheos, line 61: note: This now logs every 5 seconds, regardless of whether there was a change. If we have any reports tied to this, we may need to update them.

panentheos (approved):
See comments, then looks good!

PR #1293, opened 2023-12-21
Poll Alerts with Last Modified Header

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Rework Signs UI to direct poll for alert events](https://app.asana.com/0/1185117109217413/1205874282120116/f)

This is essentially the same change as #1291 but with the added check using the `If-Modified-Since` header. Transit data didn't have any other specific solutions for us to try on our end apart from adding tracking of this header, which they suggest would probably have prevented us from getting a stale read of alerts data.

More details on Transit data's investigation [here](https://www.notion.so/mbta-downtown-crossing/2023-11-29-Empty-alerts-response-6bc4c02795324697883d0a055a0b9d6b)


panentheos (approved):
Looks good. I mostly reviewed the last-modified code, since the rest was reviewed previously.

PR #1294, opened 2024-01-07
Revert "Poll Alerts with Last Modified Header"

PaulJKim:
Reverts mbta/signs_ui#1293

We saw the same issue occur where got a random API repose with an empty set of alerts which caused OIO settings to be cleared out.

We should put things back to the old way of doing things and think of a way to handle this case better and add this instance as another data point for the investigation on the API side of things.


PaulJKim:
Proceeding with merging and deploying this revert to prevent the OIOs settings from being wiped again.

PR #1295, opened 2024-02-20
write per-stop prediction flagging data

panentheos:
#### Summary of changes
**Asana Ticket:** [Support Concentrate consuming Signs UI headway config](https://app.asana.com/0/1185117109217413/1206385849658208/f)

This writes a new file during state saving, which contains a list of stops and whether predictions are "flagged" at that stop. This corresponds to whether headway mode was selected for a platform sign serving that stop. This computation requires the mapping from the `signs.json` file from RTS, so a copy is included. The intention is that the RTS copy is the primary record, and this copy will be manually synced when there are changes.

Note: this change depends on https://github.com/mbta/devops/pull/1572 being merged, to add the environment variable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ yarn-error.log
 /config/*.secret.exs
 test/mock_write.json
 /priv/config.json
+/priv/stops-config.json
 
 # local direnv configuration
 /.envrc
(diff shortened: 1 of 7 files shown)


PaulJKim:
Logic looks good, but I think I remember this feature being specified to being updated once per minute, regardless if there were actual updates or not. Did that change?

PaulJKim:
Ah gotcha, I should've noticed that considering I wrote the `clean` function

PaulJKim (approved):

PR #1297, opened 2024-05-23
Revert "chore: Bump gettext from 0.22.3 to 0.23.1"

PaulJKim:
Reverts mbta/signs_ui#1243


PaulJKim:
Noticed after the original Dependabot PR was merged, the deploy to dev was failing. I don't think it's related to this PR, but just in case I'm going to go ahead and revert this one and the other one I merged and revisit fixing the deploys in the AM if they're still broken.

PaulJKim:
After reverting this one, dev deploy succeeded again so leaving thing as is for now. Will revisit tomorrow morning

PR #1298, opened 2024-05-23
Revert "chore: Bump sobelow from 0.12.2 to 0.13.0"

PaulJKim:
Reverts mbta/signs_ui#1230


PaulJKim:
I didn't realize this at first, but I think deploys triggered from Dependabot fail due to something related to aws permissions so we don't need to revert this.

PR #1320, opened 2024-05-28
Shorter character limits for Amory and Babcock

PaulJKim:
#### Summary of changes
**Asana Ticket:** [Adjust Signs UI custom fields for Babcock & Amory](https://app.asana.com/0/1185117109217413/1206049304225901/f)

This PR limits custom inputs on the Amory and Babcock St signs to 16 characters per line due to the signs being shorter in width and also using a variable width font.

Due to the signs having a shorter character limit, we also decided to just exclude them from being able to be added as part of a sign group.

### DIFF
--- a/assets/css/sign_groups.css
+++ b/assets/css/sign_groups.css
@@ -70,6 +70,11 @@
   font-weight: 600;
 }
 
+.sign_groups--select_signs_row-disabled {
+  color: grey;
+  background-color: lightgray;
+}
+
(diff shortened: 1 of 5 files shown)


panentheos (approved):
Looks good, once CI passes.

PaulJKim:
@panentheos I guess `classnames` wasn't previously a dependency for Signs UI so I added it in this PR here. Is that okay with you?

PR #1337, opened 2024-06-14
Refactor Signs UI Config.State to support higher throughput

sloanelybutsurely:
### Summary of changes

- Moves `SignsUi.Config.State` into ETS (via Cachex)
- Moves scheduled config cleaning into own GenServer
- Updates how config changes are written/saved to reduce redundant writes and prevent writes from blocking reads

**Asana Ticket:** [Refactor Signs UI Config.State to support higher throughput](https://app.asana.com/0/1185117109217413/1207497364155722/f)

#### Groups of commits to hopefully make reviewing this easier:

| changes | commits |
| - | - |
| I made some small refactors `SignsUi.Config*` w/r/t code style and simplified some dependency injection. Some of these changes ended up being somewhat irrelevant as the larger refactor took place. |[`2de9a0a..1930489`](https://github.com/mbta/signs_ui/pull/1337/files/193048942a9d6626ec8d6258acde3066fd1cf454) |
| I moved/renamed `SignsUi.Config.State` to `SignsUi.Config` because the "state" portion is being moved as part of this work. I kept this as separate commits to ensure that git/GitHub would show this as a move operation and preserve as much history/blame as possible | [`74af5c8..645c4a`](https://github.com/mbta/signs_ui/pull/1337/files/d627018a7f7e8a031cfdac0fbc5723910e2bcfac..6465c4a01d273f6ba39f247a51f188bbae9d4aef#diff-135b64b77c457a0c6daa9861a4953c4e58d67b13dde0725ecfd7734ea6a1d3f1) |
| Keeping all other behavior the same, I made `SignsUi.Config` a `Supervisor` that starts a [`Cachex`](https://hexdocs.pm/cachex/Cachex.html) cache and stores all data in that cache. Functions that update state now read from the cache and then write back to it. | [`6d5043d`](https://github.com/mbta/signs_ui/pull/1337/commits/6d5043d8b295e21163854595acd5242ee0d47fbb?diff=split&w=1) |
| "Cleaning" periodically and persisting state on-change were moved into their own GenServers. Persisting state no longer blocks reads and changes that are made while state is being persisted wait and are processed "last one wins" | [`6d5043d..5fe6c4e`](https://github.com/mbta/signs_ui/pull/1337/files/6d5043d8b295e21163854595acd5242ee0d47fbb..5fe6c4e2c9fff6f089f551095876286c6a00ac72?diff=unified&w=1) |
| Refactored update/save methods to use single-key update methods when possible instead of reading-then-writing all keys. This prevents a race condition in `update_*` functions where simultaneous writes could result in data loss. | [`5fe6c4e..96292d5`](https://github.com/mbta/signs_ui/pull/1337/files/5fe6c4e2c9fff6f089f551095876286c6a00ac72..96292d53feb6b9f81e300a996e09dd9abcb81f76) |

#### Reviewer Checklist


panentheos:
I'm trying to test this locally, but it's not saving config changes. It looks like the "clean" process eventually causes the change to get persisted, but it's not getting saved immediately on editing.

sloanelybutsurely:
@panentheos i think i've fixed that in ed7fbb3. mind taking another look?

panentheos:
I'm on the fence about this. The original issue was in-process network calls blocking communication with the state module. This certainly solves that, and the code looks correct, but it's a substantial step up in complexity overall.
* I don't know that ETS itself is buying us much in this case, given how little message traffic there actually is. And doing atomic ETS updates is a bit trickier than updating process state.
* Cachex has a lot of features that we don't actually need, and having to wire it into the process tree also complicates our app.
* Debouncing the writes is neat, but adds another layer of complexity, and given that we were (mostly) getting along fine with the most naive/bad approach, I don't know if this will be a noticeable improvement over e.g. just shipping all updates off to another process via GenStage or something, so they don't block the state process.

This solution seems well put together, I'm just wondering if it will provide enough benefit given the usage characteristics of the app. Thoughts?

sloanelybutsurely:
@panentheos If you would prefer a PR with less surface area that just moves writes to storage into a GenStage consumer while maintaining the serial property of the `Config.State` process for all operations I'm happy to do that instead.

sloanelybutsurely:
Closing in favor of #1346

PR #1373, opened 2024-07-30
implement Scully compatible endpoints

panentheos:
#### Summary of changes
**Asana Ticket:** [Enable proof-of-play for new MW](https://app.asana.com/0/1205160229510941/1206767937311880/f)

This implements Scully-compatible endpoints, so messages targeting migrated SCUs show up correctly.

Note that the mapping that disambiguates multiple zones with the same name on the same SCU is not yet encoded, so this will display messages on the wrong sign in a few cases. This only affects migrated SCUs, and will be remedied before we migrate anything in production.

Also updates `signs.json` to the latest content from RTS.

### DIFF
--- a/lib/signs_ui_web/controllers/messages_controller.ex
+++ b/lib/signs_ui_web/controllers/messages_controller.ex
@@ -92,12 +92,8 @@ defmodule SignsUiWeb.MessagesController do
     with {:ok, visual_zones} <- parse_zones(conn, "visual_zones"),
          {:ok, visual_data} <- parse_visual_data(conn),
          {:ok, expiration} <- parse_expiration(conn) do
-      scu_id =
-        Enum.find_value(conn.req_headers, fn {key, value} ->
-          if(key == "x-scu-id", do: value)
-        end)
-
-      zone = MapSet.to_list(visual_zones) |> hd()
+      [scu_id] = Plug.Conn.get_req_header(conn, "x-scu-id")
+      zone = Enum.at(visual_zones, 0)
       station = SignsUi.Signs.Lookup.lookup_station_code(scu_id, zone)
       expiration_time = DateTime.utc_now() |> DateTime.add(expiration)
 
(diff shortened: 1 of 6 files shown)


lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -87,4 +87,97 @@ defmodule SignsUiWeb.MessagesController do
...
+  def background(conn, _params) do
+    with {:ok, visual_zones} <- parse_zones(conn, "visual_zones"),
+         {:ok, visual_data} <- parse_visual_data(conn),
+         {:ok, expiration} <- parse_expiration(conn) do
+      scu_id =
+        Enum.find_value(conn.req_headers, fn {key, value} ->
+          if(key == "x-scu-id", do: value)
+        end)
+
+      zone = MapSet.to_list(visual_zones) |> hd()
```
sloanelybutsurely, line 100: **quibble (non-blocking):** `Enum.at/2` would allow you to fetch a single value without iterating over the entire MapSet to convert it to a list first. Not sure how much that matters though.

```suggestion
      zone = Enum.at(visual_zones, 0)
```

lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -87,4 +87,97 @@ defmodule SignsUiWeb.MessagesController do
...
   end
+
+  def background(conn, _params) do
+    with {:ok, visual_zones} <- parse_zones(conn, "visual_zones"),
+         {:ok, visual_data} <- parse_visual_data(conn),
+         {:ok, expiration} <- parse_expiration(conn) do
+      scu_id =
+        Enum.find_value(conn.req_headers, fn {key, value} ->
+          if(key == "x-scu-id", do: value)
+        end)
```
sloanelybutsurely, line 98: **suggestion:** i think [`get_req_header/2`](https://hexdocs.pm/plug/1.16.1/Plug.Conn.html#get_req_header/2) will do this for ya

```suggestion
      [scu_id | _] = get_req_header(conn, "x-scu-id")
```

sloanelybutsurely (approved):
just a couple non-blocking stylistic things. looks good

lib/signs_ui_web/controllers/messages_controller.ex
panentheos: Ah, thanks, I didn't know about that.

PR #1433, opened 2024-10-15
Handle stations with no single-direction signs

PaulJKim:
#### Summary of changes

**Asana Ticket:** [Handle stations with no single-direction sign](https://app.asana.com/0/1185117109217413/1208171132738865/f)

Handles prediction flagging for two stations where there is only a single bidirectional sign on the platform. We could also pretty easily hardcode the few signs called out in the task, but I figured if there's a way to do it programmatically, it would be nice.

**Explaining the logic**:
At any station where there are only bi-directional signs, there would be two entries in the larger list of stop => sign mappings where the route and list of sign_ids are equal, but the stop_ids and direction_ids differ.

Therefore, the logic does a group_by in order to find groups of stop_ids/direction_ids being used at the same subset of sign_ids, and then does some filtering to check whether or not there are two directions.

### DIFF
--- a/lib/signs_ui/config/state.ex
+++ b/lib/signs_ui/config/state.ex
@@ -12,7 +12,6 @@ defmodule SignsUi.Config.State do
   alias SignsUi.Config.Utilities
 
   @sl_waterfront_route_ids ["741", "742", "743", "746"]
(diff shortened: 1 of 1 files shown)


panentheos (approved):
Took me a minute to think it through, but the logic checks out. Looks good!

PR #1481, opened 2025-02-27
Combine deploy workflows into one

PaulJKim:
#### Summary of changes
Ad-hoc

`deploy-ecs.yml` makes a prompt return

#### Reviewer Checklist

- [ ] Ask product if custom label updates in `mbta.ts` should also be made to `paess_labels.json` in Screenplay


.github/workflows/deploy-ecs.yml
```diff
@@ -1,25 +1,25 @@
...
 
 jobs:
   deploy:
-    name: Deploy to Dev
     runs-on: ubuntu-latest
     permissions:
       id-token: write
       contents: read
-    environment: dev
+    environment: ${{ github.event.inputs.environment || 'dev'}}
```
digitalcora, line 19: We should be able to remove the duplication of defaulting to `dev` in multiple places by using an environment variable, e.g.

```yml
# this goes at the top level
env:
  ENVIRONMENT: ${{ github.event.inputs.environment || 'dev'}}

# then when you want to reference it
environment: ${{ env.ENVIRONMENT }}
```

.github/workflows/deploy-ecs.yml
```diff
@@ -1,25 +1,25 @@
...
+      environment:
+        type: environment
+        required: true
+        default: dev
   push:
-    branches: master
-
-concurrency:
-  group: dev
-  cancel-in-progress: true
```
digitalcora, line 10: It should be possible to keep this and use `group: ${{ github.event.inputs.environment || 'dev' }}` (or possibly `env.ENVIRONMENT` if you take that suggestion, and assuming `env` is populated here)

.github/workflows/deploy-ecs.yml
PaulJKim: I tried this out but I'm seeing an error saying that the `env` context isn't available. I looked into it a little and unfortunately it seems that the `environment` field in a workflow is evaluated at compile time and so env variables aren't accessible :/

.github/workflows/deploy-ecs.yml
PaulJKim: Will add the concurrency stuff back in, but when trying to refernce the env variable, it seems to have the same limitation that I referenced in the other comment so it looks like we'll have to stick to referencing the inputs directly

digitalcora: Bleh. Well, I guess we can live with the duplication for now — I've been looking around but haven't found any mention of setting reusable "variables" beyond `env`, so it seems like if the context doesn't support that, you're out of luck. (Not completely out of character for GitHub Actions, sadly.)

digitalcora (approved):

PR #1483, opened 2025-03-06
display visual announcements

panentheos:
#### Summary of changes

**Asana Ticket:** [Render all PA/ESS visual messages in Signs UI](https://app.asana.com/0/1185117109217413/1209072493448524)

This modifies the sign displays to include announcements messages with a visual component. Notes:
* This is restricted to migrated stations, as well as "ad hoc" messages for non-migrated stations. The latter case mostly covers scheduled PA messages.
* The display algorithm is a best-effort approximation of what's actually playing. It doesn't take into account some finer details, including: delays from concurrent non-visual messages, priority values, delays due to other zones being busy, timing based on the audio clip length. In the future, we could improve this if desired by having the SCUs feed real-time playback info out to Signs UI or Screenplay, rather than trying to infer it.
* It includes some lightweight logic to align the client's "current time" value with the server's time, to ensure that messages show up correctly even if the browser's clock is off.


lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -177,4 +211,39 @@ defmodule SignsUiWeb.MessagesController do
...
+          {:cont, word}
+        else
+          new_acc = acc <> " " <> word
+
+          if String.length(new_acc) > max_length do
+            {:cont, acc, word}
+          else
+            {:cont, new_acc}
+          end
+        end
```
PaulJKim, line 230: Could this chunk be simplified a bit by doing something like this instead?

```
new_acc =
  if is_nil(acc),
    do: word,
    else: acc <> " " <> word

if String.length(new_acc) > max_length,
  do: {:cont, acc, word},
  else: {:cont, new_acc}

File: assets/js/SignDisplay.tsx [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+          ([innerEvents, timestamp], page) => [
+            [...innerEvents, { timestamp, value: page }],
+            timestamp + page.duration * 1000,
+          ],
+          [events, startTimestamp],
+        );
+        return [...newEvents, { timestamp: newTimestamp, value: null }];
+      },
+      [{ timestamp: 0, value: null }],
+    )
```
PaulJKim, line 41: I found this reduce a little hard to follow on without taking a few passes at it. Maybe we could break out the inner reduce out to a helper to help readability? I also found the term `Event` to be a bit ambiguous in this context and I think something slightly more descriptive could be helpful. Even just prepending `Visual` to make it `VisualEvent` or something?

lib/signs_ui_web/controllers/messages_controller.ex
panentheos: That would change the semantics in the case where a single word was longer than the limit, and cause it to emit a `nil`, so that won't work.

assets/js/SignDisplay.tsx
panentheos: I split out the inner function and decoupled the logic a little. Let me know what you think.

PaulJKim (approved):
Looks good!

PR #1490, opened 2025-07-28
fix scully zone api

panentheos:
#### Summary of changes

**Asana Ticket:** [Signs UI render for SL Chelsea inconsistent](https://app.asana.com/1/15492006741476/project/1185117109217422/task/1210880111859053?focus=true)

This modifies the API to restore compatibility with recent changes to the Scully API in how zones are specified. This fixes the display of migrated stations.

Interestingly, with the new zone format, we no longer need to look up the station code, since it's included in the compound zone name.

### DIFF
--- a/lib/signs_ui_web/controllers/messages_controller.ex
+++ b/lib/signs_ui_web/controllers/messages_controller.ex
@@ -4,7 +4,6 @@ defmodule SignsUiWeb.MessagesController do
 
   alias SignsUi.Config.SignGroups
   alias SignsUi.Messages.SignContent
-  alias SignsUi.Signs
   alias SignsUi.Signs.State
 
   plug(:laboratory_features)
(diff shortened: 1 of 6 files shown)


lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -136,16 +133,14 @@ defmodule SignsUiWeb.MessagesController do
...
          {:ok, expiration} <- parse_expiration(conn) do
-      [scu_id] = Plug.Conn.get_req_header(conn, "x-scu-id")
-
       %SignsUi.Messages.Audio{
         timestamp: DateTime.utc_now() |> DateTime.to_unix(:millisecond),
         visual_data: visual_data,
-        visual_zones: visual_zones,
-        station: Signs.Config.station_code(scu_id, Enum.at(visual_zones, 0)),
+        zones: MapSet.new(zones, &(String.split(&1, "-") |> List.last())),
+        station: Enum.at(zones, 0) |> String.split("-") |> List.first(),
```
digitalcora, line 143: It's a small thing but I wonder if we could encode the "knowledge" of this dash-separated format in one place instead of repeating it. Maybe by folding it into `parse_zones`?

digitalcora (approved):
Minor non-blocking comment, LGTM.

PR #1624, opened 2026-04-16
fix blocking alert updates

panentheos:
#### Summary of changes

This moves the alert data from process state into ETS, in order to avoid an issue where slow HTTP calls could block the process and cause readers to time out. Also uses "last modified" headers to reduce traffic slightly.

### DIFF
--- a/lib/signs_ui/alerts/state.ex
+++ b/lib/signs_ui/alerts/state.ex
@@ -19,32 +19,21 @@ defmodule SignsUi.Alerts.State do
     GenServer.start_link(__MODULE__, [], opts)
   end
 
-  @spec active_alert_ids(SignsUi.Alerts.State) :: MapSet.t(Alert.id())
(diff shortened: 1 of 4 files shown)


lib/signs_ui/alerts/state.ex
```diff
@@ -19,45 +19,37 @@ defmodule SignsUi.Alerts.State do
...
+        {:ok, :not_modified} ->
+          state
+
+        {:ok, [], _last_modified} ->
           Logger.info("empty_alerts_response: keeping current state.")
           state
 
-        {:ok, alerts} ->
+        {:ok, alerts, last_modified} ->
           new_state =
```
digitalcora: This is still called the "state", even though it's no longer part of the GenServer's state. Maybe `new_value`, since the ETS key for it is called `value`?

digitalcora (approved):
Thanks for investigating this! I was thinking all the error reports we were recently getting seemed task-worthy.

PR #1625, opened 2026-04-17
fix parsing of messages with no zones

panentheos:
#### Summary of changes

This fixes a parsing error when receiving messages with an empty zone list, which happens sometimes because of some sign configs. After the message is parsed this way, the downstream logic will correctly end up treating it as a no-op.

### DIFF
--- a/lib/signs_ui_web/controllers/messages_controller.ex
+++ b/lib/signs_ui_web/controllers/messages_controller.ex
@@ -172,8 +172,14 @@ defmodule SignsUiWeb.MessagesController do
   end
 
   defp decode_zones(zones) do
-    {Enum.at(zones, 0) |> String.split("-") |> List.first(),
-     MapSet.new(zones, &(String.split(&1, "-") |> List.last()))}
+    case Enum.at(zones, 0) do
+      nil ->
+        {nil, MapSet.new()}
+
+      zone ->
+        {String.split(zone, "-") |> List.first(),
+         MapSet.new(zones, &(String.split(&1, "-") |> List.last()))}
+    end
   end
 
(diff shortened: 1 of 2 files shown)


lib/signs_ui_web/controllers/messages_controller.ex
```diff
@@ -172,8 +172,14 @@ defmodule SignsUiWeb.MessagesController do
...
-    {Enum.at(zones, 0) |> String.split("-") |> List.first(),
-     MapSet.new(zones, &(String.split(&1, "-") |> List.last()))}
+    case Enum.at(zones, 0) do
+      nil ->
+        {nil, MapSet.new()}
+
+      zone ->
+        {String.split(zone, "-") |> List.first(),
+         MapSet.new(zones, &(String.split(&1, "-") |> List.last()))}
+    end
```
digitalcora, line 182: My clever-pattern-matching neuron is activating:

```elixir
defp decode_zones([]), do: {nil, MapSet.new()}

defp decode_zones([first | _] = zones) do
  # ...
end
```

Opinions on whether this is more readable may vary.

digitalcora (approved):

lib/signs_ui_web/controllers/messages_controller.ex
panentheos: Definitely would have done that, but the parameter is a `MapSet`, which is less amenable to matching.

digitalcora: Ohhhhhh. I guess I assumed it was a list since you were explicitly picking out the "first" item, not realizing this was just a way to pick a single _arbitrary_ item out of a set.

PR #1633, opened 2026-05-04
chore: add ueberauth_fake_oidcc in base config and test

rwaskiewicz:
#### Summary of changes

**Asana Ticket:** [Adopt ueberauth_fake_oidcc in signs_ui ](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1213899774844026?focus=true)

Introduce the `ueberauth_fake_oidcc` library to the project to be used in the dev config and test environments. Moving to this library allows us to delete the `fake.ex` Ueberauth.Strategy (and its corresponding test).

The configuration in `config.exs` has been moved to `dev.exs` to prevent dev-focused configurations making their way into other environment builds. The configuration formerly in `config.exs` predates the existence of `dev.exs`, which it is assumed is the reason it was not in `dev.exs` in the first place.

The dev remains essentially unchanged, other than options being added in order to maintain parity with the now-deleted `fake` strategy. 

~~The `auto-redirect` option has been
added to allow the "refresh-credentials" link to work as it previously did in dev, redirecting to the page served on "/" (i.e. not redirecting to a login page)~~

~~The configuration for `test` introduces `ttl` for the same reason, as well as an `auto-redirect` option so that the login form is skipped during test time.~~

The configuration for `test` introduces `auto-redirect` so that the login form is skipped during test time.

#### Reviewer Checklist

- [ ] Ask product if custom label updates in `mbta.ts` should also be made to `paess_labels.json` in Screenplay


config/config.exs
```diff
@@ -27,20 +27,6 @@ config :logger, :console,
 config :logger,
   backends: [:console, Sentry.LoggerBackend]
 
-config :ueberauth, Ueberauth,
-  providers: [
-    keycloak: {SignsUi.Ueberauth.Strategy.Fake, []}
-  ]
-
-config :ueberauth_oidcc,
```
rwaskiewicz, line 35: As stated in the PR summary, I moved this to `dev.exs`. We don't have a complimentary configuration for `ueberauth_oidcc` in `prod.exs` - does it make sense to add one? Or does the fact we r[esolve the necessary fields at runtime](https://github.com/mbta/signs_ui/blob/cf2dc019285a2fbf0348e03edbf7643420105a9c/config/runtime.exs#L28) make adding this config to `prod.exs` redundant in some way?

test/support/conn_case.ex
```diff
@@ -25,6 +25,8 @@ defmodule SignsUiWeb.ConnCase do
 
       # The default endpoint for testing
       @endpoint SignsUiWeb.Endpoint
+
+      use SignsUiWeb, :verified_routes
```
rwaskiewicz, line 29: This had to be added, as `sigil_p` was not defined for tests otherwise prior to this change

config/dev.exs
```diff
@@ -30,6 +30,27 @@ config :signs_ui, SignsUiWeb.AuthManager,
...
 
+config :ueberauth, Ueberauth,
+  providers: [
+    keycloak:
+      {Ueberauth.Strategy.FakeOidcc,
+       [
+         auto_redirect: true,
+         client_id: "dev-client",
+         roles: ["signs-ui-admin"],
+         ttl: System.system_time(:second) + 60 * 60
```
digitalcora: This probably applies to previously-committed work on other repos as well, but: Why are we using `system_time` here? The `ttl` option is an amount of time the credential is valid for (a _duration_) but `system_time` returns a _timestamp_ of seconds since epoch. The result is a credential that is valid for, at the current moment, a bit over 56 years (the amount of time between now and 1970 — Feel Old Yet?™).

The default TTL per the library's README is 9 hours, which seems fine to leave in place.

config/config.exs
panentheos: `runtime.exs` effectively runs last in the config steps, and it's providing the full configuration in prod mode, so we don't need a separate entry in `prod.exs`. Cases where it might make more sense to have a split config would be if large chunks of the configuration were environment-agnostic (`config.exs`) or not provided by env variables (`prod.exs`), but in this case most of the config comes from the environment, so it's cleaner just to set the whole thing in runtime.

panentheos (commented):
I tried running this locally but it didn't ask for roles the way Screenplay does. Is that because auto_redirect is on in `dev`?

config/dev.exs
rwaskiewicz: Ah, good call out! I goofed when reading the source of the library/made assumptions about how this duration is defined, which `iex` shows more clearly:
```elixir
iex(1)> now = System.system_time(:second)
1777987380
iex(2)> ttl = System.system_time(:second) + 360
1777987754
iex(3)> now + ttl
3555975134
iex(4)> IO.puts "Oh, oh no"
Oh, oh no
:ok
```
I'll back this out and remove it from `screens` (I don't think it applies to `screenplay`, but will double check)

rwaskiewicz:
@panentheos 

> I tried running this locally but it didn't ask for roles the way Screenplay does. Is that because auto_redirect is on in dev?

Yes, let me see if I can get that working alongside the "refresh credentials" link - right now w/o `auto-redirect`, the latter 404's.  

EDIT: Fixed in https://github.com/mbta/signs_ui/pull/1633/commits/601129466253e08cc63f88160c938cf45bdbd07e. I'll need to modify final commit message before merge, but otherwise should work now

config/dev.exs
digitalcora: Note that even if this option had been a timestamp at which the credential expires, I think this might not have done what you intended, because `config.exs` is evaluated at compile time — the credential would expire one hour after this file was last compiled, and then never be valid again until the file was changed and the server restarted.

panentheos (approved):
Looks good on my end!

PR #1649, opened 2026-06-11
allow changing custom audio TTS string

panentheos:
#### Summary of changes

**Asana Ticket:** [Custom Text Audio Modification for migrated stations](https://app.asana.com/1/15492006741476/project/1185117109217413/task/1212682780570818?focus=true)

This adds the ability to override the TTS string for custom audio messages, for migrated stations. Includes UI elements and schema changes to the config json in S3. The intended data migration strategy is as follows:
* Signs UI will handle missing or `null` `audio_text` field, treating them as `nil`.
* Saving another config _without_ updating a custom text will serialize `audio_text` as `null`.
* Saving changes to a custom text will write the given string to `audio_text`.
* RTS will handle missing or `null` `audio_text` by applying the default abbreviations, so current behavior is preserved.
* RTS will honor the new field if it's non-null, using it directly and skipping the abbreviations for migrated stations.

This should allow us to deploy normally, convert existing custom audio configs at our leisure, and then follow up with a tightening of logic/types.


assets/js/SignTextInput.tsx
```diff
@@ -4,32 +4,37 @@ function isValidText(text: string) {
...
-function SignTextInput({
-  signID,
-  line1,
-  line2,
-  onValidLine1Change,
-  onValidLine2Change,
-}: SignTextInputProps) {
-  const [showTipText, setShowTipText] = React.useState(false);
+const deriveAudioText = (value: SignTextValue) =>
+  [value.line1, value.line2].filter((x) => x).join(' ');
```
rwaskiewicz: Non-blocking - in theory, we _could_ have a bunch of whitespace in each of these text boxes, click "Modify Audio", and send a big empty string to Watts. But I doubt folks are likely to do that

rwaskiewicz (approved):

assets/js/SignTextInput.tsx
panentheos: Yeah, it's a good callout. There's also the issue of empty strings. I'm on the fence about it, given that it's an internal tool, and this "form" is extremely bare-bones, so not sure how much effort we should invest into improving it.

rwaskiewicz: No strong opinion here - I think the change would be minimal w/o affecting the output:
```diff
- [value.line1, value.line2].filter((x) => x).join(' ');
+ [value.line1, value.line2].filter((x) => x.trim()).join(' ');
```

which ought to preserve whitespace from the user input, but we don't strictly need from an AWS Polly speech quality perspective, only would prevent the rare case of sending "       ".

PR #1673, opened 2026-07-14
reduce logging level for non-critical errors

panentheos:
#### Summary of changes

This changes a few logs to use less severe levels, in order to reduce noise in our Slack alerts. The main one is the alert fetching code that sometimes encounters temporary errors or timeouts, which are actually fine as long as they don't persist. The plan is to add a new Splunk alert that will fire if we don't see a successful response for some amount of time (e.g. 1 hour).

### DIFF
--- a/lib/signs_ui_web/controllers/auth_controller.ex
+++ b/lib/signs_ui_web/controllers/auth_controller.ex
@@ -33,7 +33,7 @@ defmodule SignsUiWeb.AuthController do
         %{assigns: %{ueberauth_failure: %Ueberauth.Failure{errors: errors}}} = conn,
         _params
       ) do
-    Logger.error("ueberauth_failure #{inspect(errors)}")
+    Logger.warning("ueberauth_failure #{inspect(errors)}")
 
     if error?(errors, "bad_state") do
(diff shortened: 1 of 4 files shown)


lib/signs_ui_web/controllers/auth_controller.ex
```diff
@@ -33,7 +33,7 @@ defmodule SignsUiWeb.AuthController do
         %{assigns: %{ueberauth_failure: %Ueberauth.Failure{errors: errors}}} = conn,
         _params
       ) do
-    Logger.error("ueberauth_failure #{inspect(errors)}")
+    Logger.warning("ueberauth_failure #{inspect(errors)}")
```
rwaskiewicz, line 36: This update is causing the following test to fail, since we're capturing logs at the `:error` level still:
https://github.com/mbta/signs_ui/blob/67e12feda7b1a7bfdc321c2a4dd5b1aa1313e134/test/signs_ui_web/controllers/auth_controller_test.exs#L16-L28

rwaskiewicz (approved):
One small change needed to make the tests pass

